PR #2, opened 2017-06-19
update to dogen 2.1.0 , add dist_git plugin, remove python-setuptools, fix indent

jmtd:
This was required for CCT versions prior to 0.2, which has removed
this requirement.

PyYAML has to stay for now but can probably go when we update dogen.

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -200,7 +200,6 @@ labels:
       value: "/deployments"
 packages:
  - rh-maven33
- - python-setuptools
  - PyYAML
 # Use the run script as default since we are working as an hybrid image which can be
 # used directly to. (If we were a plain s2i image we would print the usage info here)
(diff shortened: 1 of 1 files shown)


goldmann:
This is still needed until we release new Dogen.

jmtd:
Are you sure? I tested with condev and dogen 2.0.1, and cct 0.2 zip was installed, and this PR generated and built successfully.

goldmann:
This PR https://github.com/jboss-dockerfiles/dogen/issues/148 was merged after 2.0.1 and is scheduled for 2.1.0.

jmtd:
OK I see. So the behaviour prior to that PR is to just grab the latest cct (master branch): so, it still worked since the fix in cct went into master too. But either way I don't mind waiting, especially if it means we get this fix at the same time and can stop listing PyYAML here https://github.com/jboss-dockerfiles/dogen/pull/152

PR #14, opened 2017-09-06
@WIP refactor java image to work with concreate

rcernich:
Example for new concreate tool.


image.yaml
```diff
@@ -1,241 +1,28 @@
 name: "redhat-openjdk-18/openjdk18-openshift"
 description: "Source To Image (S2I) image for Red Hat OpenShift providing OpenJDK 1.8"
-version: "1.2"
+version: "2.0"
+schema_version: 1
+from: "jboss/openjdk18-rhel7:latest"
```
goldmann: For our base images we should define the stream number, just like for OCP images. In this case it should be still `1.0`.

modules/s2i-maven/module.yaml
```diff
@@ -0,0 +1,44 @@
...
+execute:
+    - execute: install_as_root
+packages:
+    - rh-maven33
+# Use the run script as default since we are working as an hybrid image which can be
+# used directly to. (If we were a plain s2i image we would print the usage info here)
+cmd:
+    - "/usr/local/s2i/run"
+modules:
+    - name: xpaas.java
```
goldmann: Shouldn't this module have `xpaas.java.s2i-common` as dependent module?

goldmann (commented):
This would a good candidate for a first Concreate (sic!) image.

Everything looks very nice IMHO. Just a few minor comments.

Obviously later we need to think about externalizing modules to repositories so that other images could use it and we would have more control over versions. This could be next step exercise.

@rcernich What's the plan now? Should we try to keep it as WIP until we have support in Concreate for releasing images via build system? Are we OK with maintaining two separate codebases (cct_module repo and here) for now?

image.yaml
rcernich: so, 1.2, since that's what it was?  maybe 1.3, next?

modules/s2i-maven/module.yaml
rcernich: good catch

image.yaml
goldmann: We should base it on currently released version and bump version (to `1.1`) when it is necessary. Currently it's not necessary to release version `1.1` of that image because there is no feature in the development image that would be a requirements to build this image. We should bump image versions only if needed.

rcernich: You mean the image tag for the from image, not the version field, right?  If so, I think I've updated it correctly.

rcernich:
A minor update in that I defined a ``xpaas.java.feature`` module, which aggregates all the modules required for images requiring a Java runtime.  This allows other images not based on the openjdk18-openshift image to include the functionality without having to specify the individual component modules (i.e. they need to simply include ``xpaas.java.feature``).

PR #29, opened 2018-07-11
[CLOUD-2705] add usage label

jmtd:
this is to shadow rhel75's usage label which is not appropriate for this image; we point at the customer
documentation instead.

https://issues.jboss.org/browse/CLOUD-2705


rcernich (dismissed):

jmtd:
I didn't see @rcernich 's review prior to force-pushing a fix and so I don't know what it said…

rcernich (approved):

PR #30, opened 2018-07-23
Import Arquillian tests

goldmann:


jmtd (approved):
LGTM with a disclaimer: this doesn't break anything in the image, but I haven't run arq myself

tests/arq/src/test/java/org/jboss/test/arquillian/ce/openjdk/OpenJDKMysqlPersistentTest.java
```diff
@@ -0,0 +1,22 @@
...
+import org.arquillian.cube.openshift.api.OpenShiftResource;
+import org.arquillian.cube.openshift.api.Template;
+import org.arquillian.cube.openshift.api.TemplateParameter;
+import org.jboss.arquillian.junit.Arquillian;
+import org.junit.runner.RunWith;
+
+@RunWith(Arquillian.class)
+@Template(url = "file://${user.dir}/src/test/resources/openjdk18-mysql-persistent-s2i.json",
+        parameters = {
+                @TemplateParameter(name = "IMAGE_STREAM_NAMESPACE", value = "${kubernetes.namespace:openshift}")
```
rwngwn, line 12: is openshift good default here? if the xpaas images are deployed on the cluster test will run against released image. Isn't it better to fail if non-default image is not available?

tests/arq/src/test/resources/openjdk18-mysql-persistent-s2i.json
```diff
@@ -0,0 +1,639 @@
+{
```
rwngwn, line 1: should not we use templates form application templates? or how this will be synced?

rwngwn (changes requested):

tests/arq/src/test/resources/openjdk18-mysql-persistent-s2i.json
goldmann: We do provide a single application template for the openjdk image, see: https://github.com/jboss-openshift/application-templates/tree/225f292/openjdk This template is used in other test.

tests/arq/src/test/java/org/jboss/test/arquillian/ce/openjdk/OpenJDKMysqlPersistentTest.java
goldmann: This is the default that is added to the template itself: https://github.com/jboss-openshift/application-templates/blob/225f292c72603c021a650daa74b3e78d76a24c22/openjdk/openjdk18-web-basic-s2i.json#L79

rwngwn: I know.. but we probably want test to fail instead of testing wrong image and pass.

tests/arq/src/test/resources/openjdk18-mysql-persistent-s2i.json
rwngwn: ok.. this seems reasonable

tests/arq/src/test/java/org/jboss/test/arquillian/ce/openjdk/OpenJDKMysqlPersistentTest.java
goldmann: Default will not be used since `kubernetes.namespace` is set always. I can probably remove the default.

rwngwn (approved):

goldmann:
Thank you, merged!

PR #34, opened 2018-09-29
Community first!

goldmann:
What $SUBJECT says. Now, by default we build the community image, product version is built by applying the `product-overrides.yaml` file.


goldmann:
FYI @rcernich

goldmann:
@rcernich as per our chat, I'll go ahead and merge it.

rcernich:
Only comment, instead of `product-overrides.yaml`, did we want to use `rhel7-overrides.yaml`, with an eventual `rhel8-overrides.yaml` when the time comes?  I suspect we'll need to support both rhel7 and rhel8.

rcernich:
oh, one more, please update the readme with the new build instructions for community/product.

goldmann:
It was updated in subsequent commit, see the `openjdk18-dev` branch directly.

PR #39, opened 2018-11-14
Adjustments for JDK 11

goldmann:
* Renamed the OpenJDK 11 product image to `rhoar-openjdk/openjdk-11-rhel7`.
* Removed the Fedora version of the image
* Adjusted readme and build descriptor


README.adoc
```diff
@@ -1,39 +1,46 @@
...
 ```
 
 This will produce an image named `redhat-openjdk-8/openjdk18-openshift`
 
-#### JDK 11
+### RHEL JDK 11
+
 ```
-cekit --overrides-file=product-overrides.yaml --overrides-file=rhel7-jdk11-overrides.yaml build
+cekit --overrides-file=rhel7-jdk8-overrides.yaml --overrides-file=rhel7-jdk11-overrides.yaml build
```
rcernich, line 40: should we add `--redhat` flag to this?

rcernich (dismissed):
LGTM.

Do we want to use overrides for the different jdk versions, or separate image.yaml files?  Does it matter?

README.adoc
goldmann: I don't think so. This enables features specific to Red Hat environment, but should not really influence the image itself.

jmtd (dismissed):
LGTM pending rhoar

jmtd (approved):
LGTM

Issue #48, opened 2018-12-07
cekit build fails with AttributeError: 'NoneType' object has no attribute 'keys'


vorburger:
On Fedora 28:

```
dnf copr enable @cekit/cekit
sudo dnf copr enable @cekit/cekit
git clone git@github.com:jboss-container-images/openjdk.git
cd openjdk
cekit build
2018-12-07 01:45:14,980 cekit        INFO     Generating files for docker engine.
2018-12-07 01:45:14,993 cekit        INFO     Initializing image descriptor...
2018-12-07 01:45:16,894 cekit        WARNING  The way of defining repository 'jboss-os' is deprecated. Convert it to an URL based repository object. Consult Cekit docs, for more details.
Traceback (most recent call last):
  File "/usr/bin/cekit", line 11, in <module>
    load_entry_point('cekit==2.2.2', 'console_scripts', 'cekit')()
  File "/usr/lib/python3.6/site-packages/cekit/cli.py", line 289, in run
    Cekit().parse().run()
  File "/usr/lib/python3.6/site-packages/cekit/cli.py", line 223, in run
    discover_modules(os.path.join(self.args.target, 'repo'))
  File "/usr/lib/python3.6/site-packages/cekit/module.py", line 92, in discover_modules
    'module.yaml'))))
  File "/usr/lib/python3.6/site-packages/cekit/descriptor/module.py", line 31, in __init__
    self._prepare()
  File "/usr/lib/python3.6/site-packages/cekit/descriptor/image.py", line 89, in _prepare
    self._descriptor['packages'] = Packages(self._descriptor['packages'])
  File "/usr/lib/python3.6/site-packages/cekit/descriptor/packages.py", line 59, in __init__
    self._prepare()
  File "/usr/lib/python3.6/site-packages/cekit/descriptor/packages.py", line 63, in _prepare
    for x in self._descriptor.get('repositories', [])]
  File "/usr/lib/python3.6/site-packages/cekit/descriptor/packages.py", line 63, in <listcomp>
    for x in self._descriptor.get('repositories', [])]
  File "/usr/lib/python3.6/site-packages/cekit/descriptor/packages.py", line 77, in __init__
    descriptor = self._create_repo_object(descriptor)
  File "/usr/lib/python3.6/site-packages/cekit/descriptor/packages.py", line 112, in _create_repo_object
    descriptor['url']['repository'] = self._get_repo_url(descriptor)
[...]

PR #58, opened 2019-02-13
add overrides.yaml and update README (#48)

vorburger:
see #48

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -9,7 +9,7 @@ You need to https://cekit.readthedocs.io/en/develop/installation.html[install Ce
 ### CentOS JDK 8
 
 ```
-cekit build
+cekit --overrides-file artefact-overrides.yaml build
 ```
 
 This will produce an image named `centos/openjdk-8-centos7`
--- a/artefact-overrides.yaml
+++ b/artefact-overrides.yaml
@@ -1,0 +1,10 @@
+artifacts:
+- name: hawkular-javaagent
+  url: https://repo1.maven.org/maven2/org/hawkular/agent/hawkular-javaagent/1.0.1.Final/hawkular-javaagent-1.0.1.Final-shaded.jar
+  md5: 1f55173d84e4ac52d0ee542e251d8b02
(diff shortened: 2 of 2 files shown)


jmtd:
`overrides.yaml` should be renamed e.g. `artefact-overrides.yaml` to distinguish it from the other overrides YAML files. All of the overrides files should probably be documented (but that is not a blocker for this PR)

Putting this in these image sources here would be a short-term solution, because the problem is really over in [cct_modules](https://github.com/jboss-openshift/cct_module), where the artefact definitions are (e.g. https://github.com/jboss-openshift/cct_module/blob/master/jboss/container/jolokia/bash/module.yaml), as @rcernich says in #48 , we probably need community variants of the modules that are pulled in by this image, with correct checksums for the community versions.

vorburger:
> `overrides.yaml` should be renamed e.g. `artefact-overrides.yaml`

done - better?

> Putting this in these image sources here would be a short-term solution

may I open a new issue re. the long-term solution, so that this doesn't hold up merging this PR? #67

jmtd:
This is still an issue for the default `cekit build podman`, in fact any RHEL7-based image, since the prometheus and jolokia artefacts need to be in the cekit cache to build

```
CekitError: ("Error copying resource: 'jmx_prometheus_javaagent'. See logs for more info.", CekitError('Artifact jmx_prometheus_javaagent could not be found',))
```

The issue is gradually going away, the RHEL8 ones should build fine now (RPM-based artefacts), and if we get rid of Jolokia that only leaves Prometheus as the problem for the legacy images.

Perhaps image.yaml should be a RHEL8-based container, not a legacy one.

jmtd:
I'm closing this because the default images don't have the problem this addresses (RHEL8-based) and the RHEL7 ones are getting gradually less and less attention as they are effectively deprecated. I think the thrust of what this is trying to achieve is done (default images build fine in non-special environments) and that gets regularly exercised by GitHub CI too. Thanks for the work!

Issue #65, opened 2019-02-13
Improve README


vorburger:
We could improve the README of this project... I'll outline a possible future structure, for discussion.

vorburger:
https://github.com/jboss-container-images/openjdk/compare/develop...vorburger:issue-65_better-README

jmtd:
Those look like very good suggestions

vorburger:
> Those look like very good suggestions

then how about #66, as a first step?

vorburger:
and #69, as a second step?

jmtd:
Closing as the related PRs are closed, please re-open if you don't feel this has been addressed!

PR #66, opened 2019-02-13
document how to run tests in README (#65)

vorburger:
and change the overall structure by adding a level to all headings,
as a first of perhaps more improvements to the README to come.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -2,11 +2,13 @@
 
 This is the OpenShift Source to Image (S2I) image for OpenJDK.
 
+## How to develop on and contribute to it
+
 You need to https://cekit.readthedocs.io/en/develop/installation.html[install Cekit] to build these images.
 
-## Community images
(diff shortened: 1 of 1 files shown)


jmtd:
I think it would be better to document how to run `cekit test` in the cekit documentation (or improve the cekit documentation). `python3-behave` and `python3-lxml` are not the only requirements to run tests (and those package names, and the use of `dnf`, only make sense on some build host environments, e.g. RHEL8 or Fedora >?; we don't stipulate anywhere that Fedora or RHEL8 are required on the builder host, so we shouldn't implicitly require them either).

We could expand the existing "You need to install Cekit to build these images", perhaps -- to explicitly mention testing, and maybe explicitly say "see the cekit documentation for more information on building/running cekit" although the latter seems almost superfluous. But Trying to fix cekit usage problems here is the wrong place IMHO.

PR #71, opened 2019-02-14
Add content_sets_file reference in image.yaml

jmtd:
This will reduce the risk of problems if/when we need to create more branches in dist-git.


jmtd:
This doesn't seem to be necessary, I think the `--redhat` flag enables the same behaviour? Or perhaps the mere presence of `content_sets.yml`…

PR #85, opened 2019-08-22
Fix maven installation for RHEL 7

jerboaa:
Use version 3.5scl as override for the
jboss.container.maven.35.bash module as the default
for it is version 3.5. For some reason the cct_module
branch configured in image.yaml doesn't seem to matter
for branch "release". Even though it is configured to
use "sprint-35" versions get properly resolved.

Cekit mystery to me.

With this override, rh-maven35 (scl module) over plain
maven via gets installed on RHEL 7.

### DIFF
--- a/rhel7-jdk11-overrides.yaml
+++ b/rhel7-jdk11-overrides.yaml
@@ -21,6 +21,8 @@ modules:
   install:
   - name: jboss.container.openjdk.jdk
     version: "11"
+  - name: jboss.container.maven.35.bash
+    version: "3.5scl"
 
 packages:
   content_sets_file: content_sets.yml
--- a/rhel7-jdk8-overrides.yaml
+++ b/rhel7-jdk8-overrides.yaml
@@ -16,6 +16,11 @@ envs:
 - name: "JBOSS_IMAGE_VERSION"
   value: "1.6"
 
+modules:
+  install:
+  - name: jboss.container.maven.35.bash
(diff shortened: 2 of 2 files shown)


jmtd (approved):

jmtd:
Hi @jerboaa 

Thanks for this, good catch.

> For some reason the cct_module
> branch configured in image.yaml doesn't seem to matter
> for branch "release".

Yes I can't explain it. When I implemented the RHEL8 support and split the single maven cct module into "3.5" and "3.5scl", the  RHEL7 release builds seemed to pick the "right" one by default, and I hoped that this was a deterministic aspect of cekit, but it appears not. This is the appropriate fix.

jerboaa:
> Hi @jerboaa
> 
> Thanks for this, good catch.
> 
> > For some reason the cct_module
> > branch configured in image.yaml doesn't seem to matter
> > for branch "release".
> 
> Yes I can't explain it. When I implemented the RHEL8 support and split the single maven cct module into "3.5" and "3.5scl", the RHEL7 release builds seemed to pick the "right" one by default, and I hoped that this was a deterministic aspect of cekit, but it appears not. This is the appropriate fix.

Fair enough. On the other hand the branch configured for `cct_module` in `image.yaml` doesn't seem to matter. I'd have expected it to fail for rhel8 as the default for `cct_module#sprint-35` is scl. A bit too much magic happening here...

PR #87, opened 2019-08-27
Fix release branch for recent RCM changes

jmtd:
Enabling s390 for the RCM branch configuration caused the release
branch to stop building since it lacks the content_sets definitions
for the new architecture. Additionally, Cekit's module-resolution
logic has begun to pick the wrong Maven module for RHEL 7 images.

Thanks Severin Gehwolf!

### DIFF
--- a/content_sets.yml
+++ b/content_sets.yml
@@ -4,3 +4,6 @@ ppc64le:
 x86_64:
 - rhel-server-rhscl-7-rpms
 - rhel-7-server-rpms
+s390x:
+- rhel-7-for-system-z-rpms
+- rhel-7-server-for-system-z-rhscl-rpms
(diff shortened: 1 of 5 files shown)


jmtd:
Currently running an OSBS scratch build to test this.

jerboaa (commented):
This seems reasonable. For some reason the s390x seems to duplicate some of 65a831901b33df74ecc8c90aa0a6cbddf415ed14. Perhaps we should cherry-pick that from release?

jmtd:
> This seems reasonable. For some reason the s390x seems to duplicate some of 65a8319. Perhaps we should cherry-pick that from release?

Yes that was via a merge conflict resolution. It would probably look neater to cherry-pick that *first* and avoid the conflict altogether. I'll give that a go.

jmtd:
The history looks a lot neater now. @jerboaa WDYT?

jerboaa (approved):
LGTM. Thanks for fixing this. It's easier to do some archaeology with better history.

PR #92, opened 2019-09-09
remove ce.yaml: old CI config file

jmtd:
This pertained to an internal CI that has not been running for a while.


jerboaa (approved):
If it's not needed, let's remove ;-)

PR #94, opened 2019-09-17
Pin cct_module version to latest tag

jmtd:
If we're going to do the next build from the develop branch, we need
to be sure exactly which cct_module version we're getting.

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -36,7 +36,7 @@ modules:
   - name: cct_module
     git:
       url: https://github.com/jboss-openshift/cct_module.git
-      ref: master
+      ref: 0.37.0
   install:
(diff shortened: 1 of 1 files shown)


jerboaa (approved):
This is better than the current situation, so +1. We might need to change that to another tag later.

jmtd:
I've just realised that some thing I definitely wanted in the next minor release (drop hawkular) happened *after* the last tag. I'm going to check when we can get another tag in cct_module.

jmtd:
I attempted to specify the sha1-id for master as it is now instead, but cekit does not support this unfortunately https://github.com/cekit/cekit/issues/618

PR #95, opened 2019-09-19
Adjust Jolokia test to pass for future Jolokia JAR

jmtd:
This tweaked way of checking the version works with the current JAR
as well as the forthcoming 1.6.2
(see: <https://github.com/jboss-openshift/cct_module/pull/343>)


jerboaa (approved):
Looks good. Works with 1.5.0 and 1.6.2 I have locally.

PR #98, opened 2019-10-23
bump JDK8/RHEL7 minor version to reflect label update

jmtd:
Use the same technique as used in the rhel8* YAML files to avoid writing
the version string twice.


jerboaa (approved):
So we had a version mismatch.

Looks fine. This syntax is something I wasn't aware of. I've just learned something. Nice :)

jmtd:
I'm going to push a minor version bump in the same PR, otherwise CPaas might start building new micro versions that incorporate this change and we haven't shipped it

PR #99, opened 2019-10-24
remove centos, remove overrides, externalise rhel8 content_sets

jmtd:
These commits make three major changes to the structure of the Cekit YAML files

1. remove the CentOS images. We have never built or shipped them, so they are only adding complexity to the repository at present.
2. move to external content_sets file for the RHEL8 images. This is consistent with the RHEL7 images, and moves some common material to one place (so adding or removing arches for RHEL8 builds will be simpler)
3. move to complete image descriptors instead of using overrides. Overrides introduce complexity for little gain. Instead copy the (short) amount of common configuration into each YAML file. Update README to reflect new build instructions. Take the opportunity to rename the files to reflect the image names.

I've tested that the output of `cekit…osbs --sync-only` (dist-git commit) is identical before/after merging this PR, so this is a pure source-level change.


jerboaa:
This looks fine from where I'm standing. There is some duplication on the general labels, but that seems OK. I like the clarity of it using one descriptor now and not needing to look at two places (image.yml and override file).

jmtd:
Thanks. I'll pull the trigger and rebase the other in-progress branches and see how things go. We could always resurrect the centos images at a later date if we wanted them

PR #101, opened 2019-11-26
Temporarily disable incremental s2i test

jmtd:
This is broken for all flavours of image and needs investigation.

In the meantime the failure is just noise.


jerboaa (approved):
Agreed. Ship it!

PR #104, opened 2019-12-11
Track UBI8 8.1; bump minor version

jmtd:
Track UBI8 8.1 instead of 8.0.

Bump minor version since this could be a significant change.

### DIFF
--- a/openj9-11-rhel8.yaml
+++ b/openj9-11-rhel8.yaml
@@ -5,7 +5,7 @@
 
 schema_version: 1
 
-from: "ubi8:8.0-released"
+from: "ubi8:8-released"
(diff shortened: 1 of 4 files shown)


jerboaa:
I wonder whether there is a way to track latest released UBI 8 without actually hard-coding the minor version number. We'd be hitting this issue again once UBI 8.2 comes out.

jmtd:
Good point. `:8-released` seems to do what we want. Pushing a subsequent commit accordingly.

Our test suite passed for both images rebased on 8.1.

jerboaa (approved):
LGTM. Thanks!

PR #105, opened 2019-12-17
Update RHEL7-based images to Maven 3.6

jmtd:


image.yaml
```diff
@@ -42,13 +42,16 @@ modules:
   - name: cct_module
     git:
       url: https://github.com/jboss-openshift/cct_module.git
-      ref: 0.38.0
+      ref: master
```
jerboaa: I've always got a bad feeling when building from a changing target. Would we be able to add a tag to cct_module including the maven36 modules and use that instead? I'm thinking reproducible builds (at least in theory). Thoughts?

jerboaa (commented):
I'd prefer if we used a tag in cct_modules.

image.yaml
jmtd: I know what you mean. I hadn't cut a new tag in cct_modules at the time. I wonder if there's a risk that the union of all consumers of cct_modules might want tags often enough to clutter the tag namespace. On the other hand we can use git commit sha1 IDs as the ref value now (cf https://github.com/cekit/cekit/issues/618). What do you think about that?

jerboaa: That works for me. It at least fixes it to a specific revision.

jerboaa (approved):
Looks good.

PR #107, opened 2020-01-09
fix imagestreamtag openjdk-8-rhel8:1.1 image reference

gabemontero:

@jmtd - the import of imagestreamtag openjdk-8-rhel8:1.1 was failing in openshift 4.4 because of the typo here.

Can you handle creating the Jiras and linking to this PR (or creating our own Jira/PR) ?

thanks

### DIFF
--- a/templates/image-streams.json
+++ b/templates/image-streams.json
@@ -298,7 +298,7 @@
                         },
                         "from": {
                             "kind": "DockerImage",
-                            "name": "registry.redhat.io/openjdk/openjdk-1.8-rhel8:1.1"
+                            "name": "registry.redhat.io/openjdk/openjdk-8-rhel8:1.1"
                         }
                     }
                 ]


jmtd:
Hi @gabemontero , thanks for this. I've just fixed this by merging the other open PR which had e135d8558c74e1ee2d6252e25c3d8204c4d85fe7 in it.

> Can you handle creating the Jiras and linking to this PR (or creating our own Jira/PR) ?

Don't worry about that.

Aren't you going to need a new tag cutting, too? We've just bumped (most of) the image minor versions.

jmtd:
Reopening (even though it needs not be merged) since otherwise I might lose track of the conversation

gabemontero:
> Hi @gabemontero , thanks for this. I've just fixed this by merging the other open PR which had [e135d85](https://github.com/jboss-container-images/openjdk/commit/e135d8558c74e1ee2d6252e25c3d8204c4d85fe7) in it.
> 
> > Can you handle creating the Jiras and linking to this PR (or creating our own Jira/PR) ?
> 
> Don't worry about that.
> 
> Aren't you going to need a new tag cutting, too? We've just bumped (most of) the image minor versions.

Sorry yes I'll need a new tag that we can reference in https://github.com/openshift/library

jmtd:
@gabemontero 
> Sorry yes I'll need a new tag that we can reference in https://github.com/openshift/library

OK, thanks. I'd like to make another appeal for you tracking our `release` branch, rather than a new tag.

I cut a tag last time, and named it after the OCP version concerned (or tried to—4.4.0). But that hasn't changed so what should I call a new tag? The point of using a tag instead of a moving ref, from your POV (correct me if I'm wrong) is stability. But, we managed to make a screw-up in the version that was tagged last time, so that didn't help.

The policy for our release branch is: it's only merged down to when we actually ship a new image, so it always reflects the sources of the latest released image. The image streams directory within is referenced within the text we put into Erratas when we ship new images, so they are always intended to reflect the exact versions that customers should use. Thus I think it's a perfect match for you.

Compared to some other images, we are going to be changing the tagging more often. We've bumped minor versions between the ocp-4.4.0 tag and now for example, and we have at least one more bump planned by RHEL 8.2 GA. So if we go for tags we're both going to be updating things quite often.

So ­WDYT: could we try tracking `release` please?

gabemontero:
> @gabemontero
> 
> > Sorry yes I'll need a new tag that we can reference in https://github.com/openshift/library
> 
> OK, thanks. I'd like to make another appeal for you tracking our `release` branch, rather than a new tag.
> 
> I cut a tag last time, and named it after the OCP version concerned (or tried to—4.4.0). But that hasn't changed so what should I call a new tag? The point of using a tag instead of a moving ref, from your POV (correct me if I'm wrong) is stability. But, we managed to make a screw-up in the version that was tagged last time, so that didn't help.
> 
> The policy for our release branch is: it's only merged down to when we actually ship a new image, so it always reflects the sources of the latest released image. The image streams directory within is referenced within the text we put into Erratas when we ship new images, so they are always intended to reflect the exact versions that customers should use. Thus I think it's a perfect match for you.
> 
> Compared to some other images, we are going to be changing the tagging more often. We've bumped minor versions between the ocp-4.4.0 tag and now for example, and we have at least one more bump planned by RHEL 8.2 GA. So if we go for tags we're both going to be updating things quite often.
> 
> So ­WDYT: could we try tracking `release` please?

@jmtd we can use in openshift/library whatever branch / tag in your repro's you are willing to support if customer cases come in.

Feel free to open a PR against openshift/library (noting the instructions at https://github.com/openshift/library#building-the-library followed by the use of `make import` and `make verify`) to update https://github.com/openshift/library/blob/master/official.yaml#L3 to whatever you all are willing to support.

jmtd:
I've raised https://github.com/openshift/library/pull/203 so I'll close this now. Thanks!

PR #110, opened 2020-01-20
[CLOUD-3478] Add imagestreams for ppc64le and s390x

yselkowitz:


jmtd:
Hi @yselkowitz — thanks for this. A couple of quick questions before I merge

The problem this is intended to solve is to not define image versions that do not exist when running OpenShift on s390x and ppc64le. Is that right?

I assume that you've had to  create this separate file because there are no annotations/tags/etc in the imagestream specification to limit the visibilty/validity of images by architecture (which is a shame. This would be a good feature request bug)

Where/how should this file be consumed? I am aware of

1. an automatic import process that exists to read from `templates/image-streams.json` and populate things in https://github.com/openshift/library/ which is shipped alongside OpenShift releases
2. boilerplate text we copy into Erratas explaining to customers how to update their imagestream metadata from `template/image-streams.json` to get access to newly-shipped images.

Are there other avenues that consume this file/URI I am not aware of?

I don't know what code changes in Openshift or the importer might be necessary for 1). For 2), should we vary our errata boilerplate to make clear that PPC64LE/S390X users should no longer manually import from https://github.com/jboss-container-images/openjdk/blob/release/templates/image-streams.json (or they'll undo what we're trying to achieve with this separate file) and should import this new file/URI instead?

If we merge this then we are required to support this separate metadata file/URI in perpetuity I guess?

jmtd:
Another brute-force approach would for us to remove the definition of the image versions prior to adding those architectures. After all they are old and unpatched CVE images that nobody on any architecture should use any more.

I guess we might have the same issue on aarch64, I can't immediately recall when we added support for it but it wasn't from day 1. And we will cross this again when we add further architectures.

yselkowitz:
> The problem this is intended to solve is to not define image versions that do not exist when running OpenShift on s390x and ppc64le. Is that right?

Correct.  ppc64le does not have the oldest tags of openjdk18-openshift, and s390x being newer still has even fewer thereof and even of the newer images.

> I assume that you've had to create this separate file because there are no annotations/tags/etc in the imagestream specification to limit the visibilty/validity of images by architecture (which is a shame. This would be a good feature request bug)

Correct, there is no mechanism for that.

> Where/how should this file be consumed? I am aware of
> 
>     1. an automatic import process that exists to read from `templates/image-streams.json` and populate things in https://github.com/openshift/library/ which is shipped alongside OpenShift releases

Yes, these files will be added to the top-level official.yaml file in that repo, causing these files to be imported into ppc64le and s390x specific directories.  The resulting collection is then bundled in the cluster-samples-operator, which is responsible for installing them into the cluster.

>     2. boilerplate text we copy into Erratas explaining to customers how to update their imagestream metadata from `template/image-streams.json` to get access to newly-shipped images.
> 
> Are there other avenues that consume this file/URI I am not aware of?

Not that I am currently aware.

> I don't know what code changes in Openshift or the importer might be necessary for 1). 

I already have changes ready to file in PRs once this is in.

> For 2), should we vary our errata boilerplate to make clear that PPC64LE/S390X users should no longer manually import from https://github.com/jboss-container-images/openjdk/blob/release/templates/image-streams.json (or they'll undo what we're trying to achieve with this separate file) and should import this new file/URI instead?

Yes, that would be necessary.

> If we merge this then we are required to support this separate metadata file/URI in perpetuity I guess?

At least until such time that the old pre-multi-arch images are removed from the main image-streams.json file; once that occurs, we should be able to go back to a single file for all arches.  (This relates your next comment.)

> Another brute-force approach would for us to remove the definition of the image versions prior to adding those architectures. After all they are old and unpatched CVE images that nobody on any architecture should use any more.

If you would prefer that approach, I can certainly file a PR along those lines instead.  Since s390x is the newest, it would be the limiting factor, resulting in:

redhat-openjdk18-openshift: 1.7 and 1.8 only
openjdk-11-rhel7: 1.1 only
openjdk-8-rhel8: 1.0, 1.1, 1.2 (no change)
openjdk-11-rhel8: ditto
java: 8 and 11 (no change)

> I guess we might have the same issue on aarch64, I can't immediately recall when we added support for it but it wasn't from day 1. And we will cross this again when we add further architectures.

There is no openshift for aarch64 at this time, but if/when it does happen, it will definitely need a separate file, as only RHEL8 images have been released to date.  I suggest we leave that alone for now though, as I am really not sure how or when that will come to happen.

jmtd:
> If you would prefer that approach, I can certainly file a PR along those lines instead.

Thank you. I think so, but I am not 100% sure of the implications on removing the old images — I'm wondering why we've not done it already. I want to be sure that if a customer say is (still) running on one of the older images, removing them from the image stream won't break an upgrade path or similar. I'm investigating that today.

yselkowitz:
I don't know exactly what happens in that case.  The SCL imagestreams do regularly remove their EOL tags, but imagestream changes do not occur during stable releases, so that approach would first affect x86 in 4.4.

The alternative PR has been filed as #112.  EITHER this OR that needs to be merged to enable multi-arch support.

yselkowitz:
@jmtd could we get this cherry-picked to the `release` branch?  The issue atm is that the OpenJ9 changes have been merged into `develop` but they haven't been released yet, and we need to keep the openshift/library functional for P&Z.

jmtd:
> @jmtd could we get this cherry-picked to the release branch? The issue atm is that the OpenJ9 changes have been merged into develop but they haven't been released yet, and we need to keep the openshift/library functional for P&Z.

Certainly!

PR #111, opened 2020-01-20
Fix 1.8 refs in image stream metadata

jmtd:
This image name  was never valid. The :1.0 image exists in the
(correct) image stream, openjdk-8-rhel8.


jerboaa:
Looks OK.

jerboaa (approved):

PR #112, opened 2020-01-22
[CLOUD-3478] Remove obsolete tags from imagestreams

yselkowitz:
This is an alternative to #110.  EITHER this OR that needs to be committed for multi-arch support.

### DIFF
--- a/templates/image-streams.json
+++ b/templates/image-streams.json
@@ -26,146 +26,6 @@
             "spec": {
                 "tags": [
                     {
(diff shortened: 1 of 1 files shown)


jmtd:
Thank you for this! I need to do some further checking on the ramifications in OpenShift, so I've merged #110 for the time being so you can enable multi-arch, and I will revisit this once I've got a better idea what the impact may be.

PR #114, opened 2020-02-12
Update/add dependency on "singleton-jdk" module

jmtd:
This is to be merged *after* https://github.com/jboss-openshift/cct_module/pull/366

The new singleton-jdk module handles ensuring that we only have one set of JDK packages installed at a time: one version, one vendor.

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -52,6 +52,7 @@ modules:
     version: '0!3.6'
   - name: jboss.container.maven.36.bash
     version: "3.6scl"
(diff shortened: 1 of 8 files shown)


jerboaa (approved):
Seems alright. It's a bit of a mystery when an where the required env vars are being set for singleton-jdk to work.

jmtd:
They're set e.g. [here](https://github.com/jboss-openshift/cct_module/blob/master/jboss/container/openjdk/jdk/11/module.yaml#L18)

PR #116, opened 2020-02-13
[OPENJDK-59] s390x imagestreams

jmtd:
First cut, needs another eyeball

### DIFF
--- a/templates/image-streams-s390x.json
+++ b/templates/image-streams-s390x.json
@@ -15,7 +15,7 @@
             "metadata": {
                 "name": "openj9-8-rhel7",
                 "annotations": {
-                    "openshift.io/display-name": "OpenJ9 8",
+                    "openshift.io/display-name": "OpenJ9 1.8.0",
                     "openshift.io/provider-display-name": "Red Hat, Inc.",
                     "version": "1.4.17"
                 },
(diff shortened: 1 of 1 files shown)


jmtd:
Hi @yselkowitz , could you take a look at this, please? These are the OpenJ9/s390x image names/versions that we plan to ship. I wanted to merge this prior to writing Errata text, as the errata text will refer to the URI for this file. But not knowing how your import process to Openshift/library works for the architecture-specific stuff, I hoped you could give it a quick eyeball to be sure it was ok. Thanks!

jmtd:
(note to self: once this is merged, raise a develop→release merge-down PR)

templates/image-streams-s390x.json
```diff
@@ -312,27 +216,7 @@
                         },
                         "from": {
                             "kind": "DockerImage",
-                            "name": "registry.redhat.io/openjdk/openjdk-11-rhel7:latest"
-                        }
-                    },
-                    {
-                        "name": "latest",
```
yselkowitz, line 319: Why are you removing `java:latest`?

templates/image-streams-s390x.json
```diff
@@ -72,9 +52,9 @@
             "kind": "ImageStream",
             "apiVersion": "v1",
             "metadata": {
-                "name": "openjdk-11-rhel7",
+                "name": "openj9-11-rhel7",
```
yselkowitz, line 55: Couldn't `openj9-11-*` be added as alternatives to `openjdk-11-*` here, since at least the latter (unlike openjdk-8) does have a JIT for Z?  That would not contradict making openj9 the defaults for the `java` imagestream.

jmtd: Do you mean defining image streams named "openjdk-11-*" that are defined in terms of the OpenJ9 DockerImages?

(as always with OCP stuff) I'm not sure of the ramifications, and wary of setting an early precedent. Do you think that there are recipes, quickstarts etc out there that  could be potentially used without changes on Z if we did this?

PR #120, opened 2020-03-09
reset openj9*rhel8 images on release branch to be equivalent to built images

jmtd:
This reverts the sources for the openj9/rhel8 images to reflect the source state that built the current images. This targets the release branch.

Commit 9f844fb4 was over-eager. The referenced version of cct_module
does not contain the singleton-jdk module. The presently-built OpenJ9 RHEL8
images were built against the current cct_module ref, with the older
"rm-openjdk" module name.


jerboaa (approved):
LGTM.

PR #121, opened 2020-03-09
Bump cct_module ref for all images

jmtd:
This fixes build failures for some images which require the singleton-jdk
module but reference a version of cct_module prior to its introduction.

See also GitHub PR #120.

Here I take the opportunity to harmonize the cct_module ref across all the
development images.


jerboaa (approved):
OK.

PR #124, opened 2020-04-01
Update cct_module for maven module changes

jmtd:
module simplifications in cct_module

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -41,17 +41,18 @@ modules:
   - name: cct_module
     git:
       url: https://github.com/jboss-openshift/cct_module.git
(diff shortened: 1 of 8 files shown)


jmtd:
The version pinning for the prometheus module is "7" for all the flavours of images, even the RHEL8 ones, because the alternative version (8.2) relies upon RPMs in RHEL 8.2 which is not GA yet, so the public rhel8-based images are not yet using that as a base, and need to use the older scheme.

jerboaa (approved):
I suppose this is ok (note that I have not been following latest developments in cct_modules).

PR #127, opened 2020-05-04
Re-enable singleton-jdk for ubi8/openjdk-11

jmtd:
Jiri A noticed that java-1.8.0-openjdk-headless had snuck back into the ubi8/openjdk-11 RC build, and so this module is sadly still necessary for now.


jmtd:
I'm going to just merge this immediately to get this out fast. On my head be it!

PR #131, opened 2020-05-19
Fix incremental S2I tests

jmtd:
The exact string output by Maven had changed since the test was
originally written. Retag the test to run for all the OpenJDK
images.


jerboaa (approved):
LGTM

PR #137, opened 2020-07-16
Split RHEL7 and UBI8 imagestreams

yselkowitz:
This is needed in order to move forward with RHEL8-only content in OCP.

Signed-off-by: Yaakov Selkowitz <yselkowi@redhat.com>

/cc @jmtd


yselkowitz:
We have to take a different approach to this.

PR #142, opened 2020-08-12
[CLOUD-3762] Restructure java imagestreams

yselkowitz:
Following the model being implemented by the SCLs, imagestreams are now
defined by both runtime and underlying OS version, with an emphasis on
making UBI 8 available and default.

In the case of Java, we also have both OpenJ9 alongside OpenJDK on some
architectures as an additional option.  The naming scheme used here
should cover any expected permutations going forward.

### DIFF
--- a/templates/community-image-streams.json
+++ b/templates/community-image-streams.json
@@ -90,9 +90,9 @@
             "spec": {
                 "tags": [
                     {
-                        "name": "8",
+                        "name": "openjdk-8-ubi8",
                         "annotations": {
-                            "openshift.io/display-name": "Red Hat OpenJDK 8",
+                            "openshift.io/display-name": "Red Hat OpenJDK 8 (UBI 8)",
                             "description": "Build and run Java applications using Maven and OpenJDK 1.8 upon UBI8.",
                             "iconClass": "icon-rh-openjdk",
                             "tags": "builder,java,openjdk",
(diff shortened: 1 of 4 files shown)


jmtd:
Hi @yselkowitz thanks for this, it looks good. I think as soon as I merge this OKD will pick it up so we could see what impact this has on the generation scripts at openshift/library for OKD, and/or in a cluster setup too, before cherry-picking over to release. Depending on urgency.

One question, the form of the imagestream names you've chosen is e.g. "openjdk-8-ubi8", whereas our image names are the other way around, ubi8/openjdk-8 or ubi8-openjdk-8.yaml. Is the form you've chosen to be consistent with what SCL are doing?

jmtd:
> One question, the form of the imagestream names you've chosen is e.g. "openjdk-8-ubi8", whereas our image names are the other way around, ubi8/openjdk-8 or ubi8-openjdk-8.yaml. Is the form you've chosen to be consistent with what SCL are doing?
I'm merging (to develop) ahead of hearing back from you on this

yselkowitz:
Thanks @jmtd this is the impact on OKD:

https://github.com/openshift/library/commit/ad744b03994a7f056782f8e2800353baac9eb86a

The SCLs are moving to a naming scheme of e.g. nodejs:12-ubi8.  Java is in a unique position where multiple implementations exist on some architectures.  To keep a common set of imagestream tags across architectures, I chose to prefix those with the implementation name, e.g. java:openjdk-8-ubi8.  Hope that helps make sense of this renaming.

PR #143, opened 2020-08-13
[OPENJDK-158] Update from: for RHEL8/UBI8-based images

jmtd:
Internal support for ":*-released" tags is going away by the end
of September. Meanwhile, internal support for a fully-qualified
external base image URI is now available.

Update the from: for RHEL8-based images to a fully qualified,
globally available container registry URI.

This also makes simple local building of those images easier.

### DIFF
--- a/openj9-8-rhel8.yaml
+++ b/openj9-8-rhel8.yaml
@@ -4,7 +4,7 @@
 
 schema_version: 1
 
-from: "ubi8-minimal:8-released"
+from: "registry.access.redhat.com/ubi8/ubi-minimal"
 name: &name "openj9/openj9-8-rhel8"
 version: &version "1.2"
 description: "Source To Image (S2I) image for Red Hat OpenShift providing OpenJ9 1.8"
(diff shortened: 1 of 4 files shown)


openj9-8-rhel8.yaml
```diff
@@ -4,7 +4,7 @@
 
 schema_version: 1
 
-from: "ubi8-minimal:8-released"
+from: "registry.access.redhat.com/ubi8/ubi-minimal:latest"
```
jerboaa: I think you can drop the `latest` tag. Just using `registry.access.redhat.com/ubi8/ubi-minimal` should be fine.

ubi8-openjdk-11.yaml
```diff
@@ -2,7 +2,7 @@
 
 schema_version: 1
 
-from: "ubi8-minimal:8-released"
+from: "registry.access.redhat.com/ubi8/ubi-minimal:latest"
```
jerboaa: Same. You shouldn't need the latest tag.

jerboaa (changes requested):
This is a good idea. I needed to do this in another image too. I'd suggest to omit the `:latest` tag suffix, though.

PR #144, opened 2020-08-13
Updated openj9 rhel7 content sets list for s390x

mdrafiur:

- [X] Pull Request contains link to the JIRA issue

Jira Issue:
https://issues.redhat.com/browse/OPENJDK-146

### DIFF
--- a/content_sets_rhel7.yml
+++ b/content_sets_rhel7.yml
@@ -1,5 +1,6 @@
 s390x:
 - rhel-7-for-system-z-rpms
+- rhel-7-for-system-z-optional-rpms
 - rhel-7-server-for-system-z-rhscl-rpms
 ppc64le:
 - rhel-7-server-for-power-le-rhscl-rpms


jerboaa:
I'm not sure this is the right fix. Please investigate an alternative fix to spec file requires.

yselkowitz:
I don't see an alternative here.  Unlike the RHEL OpenJDK packages, the OpenJ9 packages are not rebuilt from source, so the patch in the former that allows depending on pcsc-lite-libs instead of -devel can't be applied to the latter.  As such, the patch is okay but incomplete; there is no reason not to fix ppc64le at the same time.

jerboaa:
> I don't see an alternative here. Unlike the RHEL OpenJDK packages, the OpenJ9 packages are not rebuilt from source, so the patch in the former that allows depending on pcsc-lite-libs instead of -devel can't be applied to the latter. As such, the patch is okay but incomplete; there is no reason not to fix ppc64le at the same time.

@yselkowitz Hmm, good point. We should upstream this patch:
https://git.centos.org/rpms/java-1.8.0-openjdk/blob/c7/f/SOURCES/rh1684077-openjdk_should_depend_on_pcsc-lite-libs_instead_of_pcsc-lite-devel.patch

yselkowitz:
The problem with upstreaming that patch is what happens if/when the SONAME changes (off of ABI version 1)?  Or other *NIX platforms which don't use the same naming scheme (unless this code is only used on Linux)?  This is all a bit tricky.

I suppose the alternative would be to get the unversioned .so shipped in -libs instead of -devel.  That happens *occasionally* but not very often.

levivic:
An commit for openj9 is at http://pkgs.devel.redhat.com/cgit/rpms/java-1.8.0-openj9/commit/?h=openj9-1-rhel-7&id=c3dced6a7fe53ad033f3b7a1432bcbdb2217580d. Will test if it works as suggested by @jerboaa in https://issues.redhat.com/browse/OPENJDK-146?focusedCommentId=14410356&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14410356

levivic:
Update: just confirmed, without the patch https://git.centos.org/rpms/java-1.8.0-openjdk/blob/c7/f/SOURCES/rh1684077-openjdk_should_depend_on_pcsc-lite-libs_instead_of_pcsc-lite-devel.patch being applied to openj9 rpm, the bug still exist. So the change this PR made seems the only solution to fix the content_set failure for now.

jerboaa (approved):
What @levivic said. It seems the only option to fix the content set issue short term. OK with me. Thanks for baring with me.

yselkowitz:
Please fix ppc64le at the same time.

mdrafiur:
@yselkowitz @jerboaa Local Branch was deleted accidentally and therefore submitted a new PR #148 with the changes suggested. Sorry for the inconvenience. 

cc @levivic

jerboaa:
@mdrafiur OK closing this in favor of #148

PR #145, opened 2020-08-20
[CLOUD-2137] Add readiness probe to OpenJDK template

jmtd:
Ported version of https://github.com/jboss-openshift/application-templates/pull/392


jmtd:
This improves the ability for OpenShift to determine whether a pod is ready or not: if it can establish a TCP connection to port 8080, it's considered live. A pod starting up or shutting down (or broken) will not show as live in the UI if this probe fails. I've just tested loading this amended template into an OCP cluster and it was fine.

jerboaa (approved):
LGTM

PR #149, opened 2020-09-02
[OPENJDK-159] fix FROM lines for OpenJ9 images

jmtd:
registry.access.redhat.com is not a permitted FROM source in OSBS.
Use registry.redhat.io instead, which is (but requires authentication)


jerboaa (approved):
OK

PR #151, opened 2020-09-07
Openjdk 37 drop jolokia

jmtd:
https://issues.redhat.com/browse/OPENJDK-37

### DIFF
--- a/openjdk-11-rhel7.yaml
+++ b/openjdk-11-rhel7.yaml
@@ -44,8 +44,6 @@ modules:
     version: "11"
   - name: jboss.container.prometheus
     version: "7"
-  - name: jboss.container.jolokia
-    version: "7"
   - name: jboss.container.java.s2i.bash
   - name: jboss.container.maven
     version: "7.0.3.6"
--- a/openjdk18-openshift.yaml
+++ b/openjdk18-openshift.yaml
@@ -45,8 +45,6 @@ modules:
     version: "8"
   - name: jboss.container.prometheus
     version: "7"
-  - name: jboss.container.jolokia
-    version: "7"
   - name: jboss.container.java.s2i.bash
   - name: jboss.container.maven
     version: "7.0.3.6"
--- a/templates/openjdk-web-basic-s2i.json
+++ b/templates/openjdk-web-basic-s2i.json
@@ -260,11 +260,6 @@
                                 ],
                                 "ports": [
                                     {
-                                        "name": "jolokia",
-                                        "containerPort": 8778,
-                                        "protocol": "TCP"
-                                    },
-                                    {
                                         "name": "http",
                                         "containerPort": 8080,
                                         "protocol": "TCP"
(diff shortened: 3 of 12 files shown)


jmtd:
- [ ] We need this too https://github.com/jboss-openshift/cct_module/pull/383/files
- [x] and  to confirm that OpenJ9 want this dropped in develop

image.yaml
```diff
@@ -32,6 +32,7 @@ envs:
 ports:
 - value: 8080
 - value: 8443
+- value: 8778
```
jerboaa: This definitely is in need of a comment. FWIW, I'm not sure why this cannot be enabled by images which add jolokia themselves. Isn't the idea for them to use `FROM openjdk/openjdk-11` for example and then add it back in? Pretty sure this can be achieved with a docker build strategy on OpenShift with subsequent s2i customizations.

jerboaa (changes requested):
I'm not convinced we should expose ports which are only optionally used. This essentially ties us back to some having-jolokia-zombie-mode which doesn't seem a good idea.

image.yaml
jmtd: I think it is possible with a docker build strategy. I didn't think it was possible with a "pure" s2i build (the kind which we exercise when we run the s2i binary outside of openshift) but I might be out-of-date wrt with what it can do. I guess I'm happy to close the port so long as we can document (in the developer's blog post perhaps) how to open it again. I'll take a look at that.

jerboaa: Thanks!

jmtd: the current draft blog post points at this (WIP) solution for re-activating Jolokia. It's a pure S2I-based approach at the moment, requiring a custom assemble script and a custom environment https://github.com/jmtd/quarkus-quickstarts/commit/7f9e866aa0dab735ec8d7397f78b40cad46745b5
I don't *think* its possible with just the above to open an additional port,  so I'll double-check that and see how much more complicated it would be to do this another way (Docker build strategy in OpenShift for instance. I have an abandoned blog post about something else that touched on that I might be able to re-use)

jerboaa: Understood that s2i might not work, but that's expected (to me). After all the use case would not to build application source with it, but to extend our image with new libraries. A custom docker build strategy for this would make more sense. Users would then use the resulting s2i image to build their apps with jolokia.

jerboaa (commented):
Why the difference between openj9 yaml files and openjdk yaml files? Perhaps we shouldn't touch openj9 or do the same for them than we did with openjdk yamls?

jmtd:
> Why the difference between openj9 yaml files and openjdk yaml files? Perhaps we shouldn't touch openj9 or do the same for them than we did with openjdk yamls?

You're right, I am just revisiting this PR again, I was in touch with the OpenJ9 team a week or so ago and proposed I don't touch their images any more, so I'll remove those bits from this PR.

Thanks

jmtd:
@jerboaa I've re-done this now and I think it's ready for review again. TIA!

jerboaa (approved):
OK

PR #153, opened 2020-09-14
Initial Github Action/workflow for building

jmtd:
This adds two GitHub workflows (one for JDK 8, one for JDK 11) that will trigger on every pull request or commit. The two workflows attempt to build the UBI8 images (via cekit 3.6.0 and the docker back-end) and will report if that fails.

Future improvements

 * parallel build the different JDK flavours
 * speed up the build process by pre-preparing a cekit builder?
 * auto publish the build containers somewhere in some cases


jerboaa (approved):
Looks good to me. Tests seem fine here:

https://github.com/jmtd/openjdk/runs/1111592839?check_suite_focus=true

PR #157, opened 2020-09-29
Pin base images to RHEL minor version

jmtd:
After thinking about it for a while, I think we should explicitly stick
to a given RHEL minor release as our FROM image, and manage an upgrade
to the new minor release when available, bumping our own minor version
at the same time. Rationale: a given RPM might have a function or ABI
change in a RHEL minor release (I believe), which could cause behaviour
changes in our images, so we should gate those changes with a minor
version bump ourselves.

This is necessary for the RHEL7 bases too, as current minor is 7.8 and
there's a 7.9 in the pipeline.

Depending on scheduling we may want  to cherry-pick this to the release
branch too, but it might not be necessary.

### DIFF
--- a/.github/workflows/ubi8-openjdk-11.yml
+++ b/.github/workflows/ubi8-openjdk-11.yml
@@ -3,6 +3,7 @@ on: [push, pull_request]
 env:
   LANG: en_US.UTF-8
   CEKIT_VERSION: 3.6.0
+  BASE_IMG_VERSION: 8.2
 jobs:
   openjdkci:
     name: OpenJDK S2I Build and Test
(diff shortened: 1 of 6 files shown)


jerboaa:
> Rationale: a given RPM might have a function or ABI
> change in a RHEL minor release (I believe), which could cause behaviour
> changes in our images, so we should gate those changes with a minor
> version bump ourselves.

This seems misguided:
https://access.redhat.com/solutions/5154

The promise for RHEL is ABI compatibility throughout a major RHEL release.

jerboaa (changes requested):
Sorry, but this is big -1

jmtd:
Hi @jerboaa: I originally opened this whilst mixing up some issues we've had with Maven with issues we've had with Prometheus (and snakeyaml). I realised that, without this, the UBI8 images will automatically update their prometheus agent once 8.3 is GA. I mistakenly thought that the agent had gone up a minor version between RHEL 8.2 and 8.3 (I was mixing it up with Maven 3.5 → Maven 3.6) and we should gate that change in our images with a corresponding minor version bump.

I *also* was mistaken about the snakeyaml version bump in the Prometheus agent: I thought it was from 1.16 → 1.26, which would very likely have some fairly significant changes, but a) they might not be reflected in the Prometheus interfaces anyway,  and b) it's actually only 1.25 → 1.26, the 1.16 was an out-of-date reference and I see you've fixed that in the RPMs already.

Ultimately the agent JAR has only bumped a micro version. I haven't gone over their versioning policy with a fine toothed comb but that's probably fine for us to include in a respin after all. So, there's no need for this PR.

Thanks for your feedback!

PR #159, opened 2020-10-06
Default ubi8 public builds

jmtd:
Clean up the documentation to document building outside a redhat environment, prioritise the UBI8-based images, provide redhat-specific information in the form of override YAML files, fix the catalog links, etc.


jerboaa (approved):
Seems OK to me.

PR #162, opened 2020-10-12
Inline the container.yaml file

jmtd:
In a similar vein to the content-sets files, inline the OSBS
configuration "container.yaml" for the RHEL7 images (and fix
their building)


jerboaa (approved):
OK

PR #173, opened 2020-12-18
Add OpenJDK packages explicitly to ODCS Composes

jmtd:
This allows us to build containers with OpenJDK packages that have not
yet shipped in RHEL but are about to; we can then ship the container and
the OpenJDK RPM update at the same time.

Thanks Severin for the suggestion.


jerboaa (approved):
LGTM

PR #176, opened 2020-12-23
Bump cct_module to 0.44.0

jmtd:
This pulls in
"blank maven.conf instead of removing it"
<https://github.com/jboss-openshift/cct_module/pull/391>
which should fix the Behave tests in GitHub Actions CI, and

"Enable maven AppStream module via conf file"
https://github.com/jboss-openshift/cct_module/pull/390


jerboaa (approved):
OK

PR #180, opened 2021-01-19
[redhat] Use signing-intent release

jerboaa:
We should use signing intent release for the ODCS packages. This has the advantage
that the build fails with unsigned packages.


jerboaa:
@jmtd Please review!

jmtd (approved):
LGTM

PR #181, opened 2021-02-10
redhat metadata fixup: don't inherit content sets, ubi8/jdk8 microdnf

jmtd:
* re-order packages/osbs stanza in RHEL7/JDK8 file to match the others
 *  remove inherit: true from ubi8 images to prevent leaking pre-GA content
 * re-add manager: microdnf to ubi8/jdk8 that was accidentaly dropped in an earlier refactoring

### DIFF
--- a/redhat/ubi8-openjdk-11.yaml
+++ b/redhat/ubi8-openjdk-11.yaml
@@ -3,7 +3,6 @@ osbs:
     container:
       compose:
         pulp_repos: true
-        inherit: true
         packages:
         - java-11-openjdk
         - java-11-openjdk-devel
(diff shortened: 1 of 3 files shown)


redhat/ubi8-openjdk-11.yaml
```diff
@@ -3,7 +3,6 @@ osbs:
     container:
       compose:
         pulp_repos: true
-        inherit: true
```
jerboaa, line 6: It's not clear what effect that will have for building the container against a pre-GA OpenJDK package. However, let's remove and add back should we need it in future.

jerboaa (approved):
OK. If we need to set `inherit: true` later, we'll have it in the back of our heads ;-)

PR #187, opened 2021-02-22
Adjust test-tagging to account for runtime images

jmtd:
Any test tagged '@ubi8' will run on the new runtime image descriptors,
but they are all currently dependent upon performing S2I builds which
is not supported.

Adjust the tagging to be specific to the ubi8/openjdk-8 and
ubi8/openjdk-11 images.


jerboaa (approved):
Looks fine.

PR #199, opened 2021-04-06
[OPENJDK-312] explicitly depend upon "which"

jmtd:
run-java.sh calls "which" in some circumstances but does not explicitly
depend upon it.


jerboaa (approved):
OK. An alternative could be to use `type -p` throughout, a bash built-in.

jmtd:
> OK. An alternative could be to use `type -p` throughout, a bash built-in.

Oh that's even better! Thanks. PR updated.

PR #202, opened 2021-04-16
[OPENJDK-131] Review container maxMetaspaceSize tuning

Josh-Matsuoka:
This PR addresses https://issues.redhat.com/browse/OPENJDK-131 , from reading through the linked issues it is unnecessary to set the default max metaspace. Currently we set it by default to 100m if no max metaspace argument was passed, this PR removes that and defaults to not passing the argument at all. The tests have been adjusted accordingly.


jerboaa (approved):
This looks good.

In times where the JVM itself is aware of container limits, this should no longer be necessary and 100m by default seems unrealistic for a one-size-fits-all.

I'll let @jmtd have the last word as he's more aware of the legacy reason(s) this is there in the first place.

PR #204, opened 2021-04-22
Bump and harmonize image versions to 1.9

jmtd:
For simplicity's sake I'd like to harmonize all the image versions to the same number. The oldest/highest shipped image is RHEL7/JDK8 with version 1.8, so I moved them all up to 1.9.


jerboaa (approved):
LGTM

PR #205, opened 2021-04-22
[OPENJDK-37] Delay removing Jolokia to the next release

jmtd:
I'm very keen to do this, but I think it's worthy of reconsideration in light of one new development, and either way we need to make sure we document this well, and we aren't in a position to get docs/blogs out about it quickly right now, but I do not want to delay the other pending stuff in `develop`. So I propose we drop this for 1.9, ship 1.9, then do this for 2.0.

https://issues.redhat.com/browse/OPENJDK-37

### DIFF
--- a/openjdk-11-rhel7.yaml
+++ b/openjdk-11-rhel7.yaml
@@ -41,6 +41,8 @@ modules:
     version: "11"
   - name: jboss.container.prometheus
     version: "7"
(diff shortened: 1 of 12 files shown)


jerboaa (approved):
OK. We should make sure to file an issue to get it removed (again) in the next release.

PR #207, opened 2021-04-22
Added support for Quarkus build layouts

### LINKED ISSUE #206: Make run-java.sh support Quarkus
It would be useful if the script would by default understand Quarkus apps.

I think the "uber-jar" build layout already works (but it would have to be tested to be sure), but the other two layouts are not. The "legacy-jar" one always creates two jars (one xxx.jar and one xxx-runner.jar, it's the latter that's needed) plus a lib folder, while the new "fast-jar" option creates an entire folder with a bunch of files (including the dependencies in a lib folder and even a third jar with the fixed name quarkus-app.jar).

quintesse:
Quarkus, in its latest version, has 3 different build layouts that
it can create. They are called "legacy-jar", "uber-jar" and the
nwest is called "fast-jar". Of these 3 `run-java.sh` only supports
the "uber-jar". This commit adds support for the other two.

Fixes #206

    Sorry, no idea that the proper tag should be

    Sorry, no JIRA issue, just a GH issue

### DIFF
--- a/modules/run/bash/artifacts/opt/jboss/container/java/run/run-java.sh
+++ b/modules/run/bash/artifacts/opt/jboss/container/java/run/run-java.sh
@@ -35,13 +35,23 @@ auto_detect_jar_file() {
   local old_dir=$(pwd)
   cd ${dir}
   if [ $? = 0 ]; then
-    local nr_jars=`ls *.jar 2>/dev/null | grep -v '^original-' | wc -l | tr -d '[[:space:]]'`
-    if [ ${nr_jars} = 1 ]; then
-      ls *.jar | grep -v '^original-'
-      exit 0
-    fi
+    if [ -f "quarkus-app/quarkus-run.jar" ]; then
+        echo quarkus-app/quarkus-run.jar
+    else
+        local nr_jars=`ls *.jar 2>/dev/null | grep -v '^original-' | wc -l | tr -d '[[:space:]]'`
+        if [ ${nr_jars} = 1 ]; then
+            ls *.jar | grep -v '^original-'
+            exit 0
+        fi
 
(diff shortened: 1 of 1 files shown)


jmtd:
This looks good to me, I think we need some test coverage. I'm just looking at how simple this would be to add.

quintesse:
Yeah the tests are probably the actual hard part of this issue 😅

quintesse:
Btw a Quarkus app can be compiled with the different build layouts using the option `-Dquarkus.package.type=(legacy-jar|uber-jar|fast-jar)` (where fast-jar is the default).

jmtd:
that's exactly what I needed, thanks --- I've got a very basic test in the works. I'll raise it as as PR against your PR branch (I think that works)

jmtd:
I spent the morning thinking about this and poking at the problem a little. I'm just running the whole test suite over your branch to see what happens with other (non-quarkus) quickstarts, although I don't expect problems.

The quarkus people fixed the getting-started quickstart by adding a bunch of s2i environment variables to point run-java at the right place for the quarkus jar: https://github.com/quarkusio/quarkus-quickstarts/commit/dfe99e36cec416ca7a60031b53973975086e9638

I wrote some tests to build that with the legacy-jar and fast-jar build strategies. But of course, they both pass now even before this commit due to that change in the quickstart repo...

I think there's definitely an argument to be made that having quarkus work OOTB with this image, without requiring the user to specify things in .s2i/environment is a good idea.

quintesse:
Indeed, there's no reason why people would be using S2I, that's very OpnShift-centric

jmtd:
Subsumed into #358 . Thanks for your contribution and sorry for the delay!

PR #208, opened 2021-04-27
[CLOUD-3977] Deleting src dir after s2i breaks WildFly image with ODO

jmtd:
This is a cherry pick of a commit by JF Denise over in cct_modules.
Whilst working on #207 I realised this was useful for diagnosing what
is going on when builds are not working as expected.


jerboaa (approved):
OK

PR #212, opened 2021-05-04
Add Runtime ImageStream metadata

jmtd:
I've verified this is syntactically and semantically OK by loading it into an OCP 4.7 test cluster. However the image URIs are necessarily not correct until the runtime images are GA.

I'm putting them in a totally distinct JSON file so that they are not auto-pulled into OCP's default imagestream metadata until we can review and consider *all* our imagestream metadata with respect to how they are represented and interact in the OCP UI. Since these aren't builder images they shouldn't show up as quickstarts, some DIY is needed to use these.

We can reference this JSON file in the Errata text.

### DIFF
--- a/templates/runtime-image-streams.json
+++ b/templates/runtime-image-streams.json
@@ -1,0 +1,79 @@
+{
+    "kind": "List",
+    "apiVersion": "v1",
+    "metadata": {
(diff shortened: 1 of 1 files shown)


jerboaa (approved):
Seems OK to me.

PR #213, opened 2021-05-06
Add Runtime ImageStream metadata

jmtd:
This is a cherry-pick over to the `release` branch.

I'd like to move away from the `release` branch starting from the next
actual container `release` (1.9). My intention is to get `develop` into
the right state, build, and when we ship, tag the `develop` branch at
that point with `1.9`. From that point forward we would work entirely
from the `develop` branch.

A chicken-and-egg issue here is I need a stable URI to reference in the
errata to ship the runtime images, and I'd rather not use the `develop`
branch. We currently use the `release` branch for the latest correct
metadata for OpenShift, so it seems the appropriate place for this too
for now.

### DIFF
--- a/templates/runtime-image-streams.json
+++ b/templates/runtime-image-streams.json
@@ -1,0 +1,79 @@
+{
+    "kind": "List",
+    "apiVersion": "v1",
+    "metadata": {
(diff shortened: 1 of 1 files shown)


jerboaa (approved):
Sure.

PR #214, opened 2021-05-20
[OPENJDK-390] fix JAVA_HOME for runtime images

jmtd:
The path /usr/lib/jvm/jre exists in both images, the prior
values do not.

https://issues.redhat.com/browse/OPENJDK-390

### DIFF
--- a/modules/jre/11/module.yaml
+++ b/modules/jre/11/module.yaml
@@ -14,7 +14,7 @@ labels:
 
 envs:
 - name: "JAVA_HOME"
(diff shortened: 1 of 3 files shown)


jerboaa (changes requested):
Ouch. There should be a behave test for it. Something like `$JAVA_HOME/bin/java -version` should be sufficient.

jmtd:
@jerboaa agreed!

jerboaa (approved):
LGTM

PR #218, opened 2021-06-14
[OPENJDK-407]: clean up RPMs that are not in the UBI set

jmtd:
The new cct_module version includes the updated maven modules which enable the relevant dnf module without requiring full dnf to be installed. This reduces the number of weak-deps which get pulled in and not removed afterwards. Additionally, the new jboss.container.util.ubi-clean module removes a static list of RPMs known to not be in the UBI set.

https://issues.redhat.com/browse/OPENJDK-407


jerboaa (approved):
LGTM

PR #224, opened 2021-06-25
Stop referencing Image :latest tags

jmtd:
Substitute every reference to an OpenJDK Image (not ImageStream) :latest
tag with the current latest major.minor version instead.

ImageStream tag definitions are not modified, so where ":latest"
ImageStream tags already exist, they should continue to do so, just
referring to the back-end Images by different tags.

For OpenShift end-users, there should be no impact, as the ImageStreams
themselves retain the same tags pointing at (ultimately) the same
images.

This is in preparation for moving the OpenJDK images to "multi-stream",
a consequence of which is the image ":latest" tags will stop being
published in the Container Catalog. I'm pushing it now, to give it a
chance to percolate through OpenShift installations ahead of that
switch.

This does not impact the OpenJ9 images.

### DIFF
--- a/templates/community-image-streams.json
+++ b/templates/community-image-streams.json
@@ -106,7 +106,7 @@
                         },
                         "from": {
                             "kind": "DockerImage",
-                            "name": "registry.access.redhat.com/ubi8/openjdk-8:latest"
+                            "name": "registry.access.redhat.com/ubi8/openjdk-8:1.3"
                         }
                     },
                     {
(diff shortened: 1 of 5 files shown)


jmtd:
Hi @yselkowitz - it's been a while since we talked! Would you be OK to give this a look-over, since I've touched the multi-arch ImageStream definitions?

yselkowitz:
@jmtd it has indeed been a while!  Sorry for the delay, been busy with other things.  I'm not quite understanding what you're trying to achieve here, so maybe best if we make a time to talk rather than a bunch of back and forth on a PR.

jmtd:
Sure, I'll mail you to schedule a chat.

jmtd:
* Out of date wrt the latest image versions
 *  Not clear from PR what the driver for this is

jmtd:
Ah I remember, this was prerequisite work for moving to multi-stream images, which we have not done and probably won't do.

PR #227, opened 2021-07-12
[OPENJDK-111] Adjust JVM parameters when JAVA_DIAGNOSTICS is set

jmtd:
This is for the situation where JAVA_DIAGNOSTICS=true

<https://issues.redhat.com/browse/OPENJDK-111>

### DIFF
--- a/modules/jdk/11/artifacts/opt/jboss/container/openjdk/jdk/jvm-options
+++ b/modules/jdk/11/artifacts/opt/jboss/container/openjdk/jdk/jvm-options
@@ -6,5 +6,5 @@
 # ==============================================================================
 
 function jvm_specific_diagnostics() {
-    echo "-Xlog:gc::utctime -XX:NativeMemoryTracking=summary"
+    echo "-Xlog:gc::utctime -XX:NativeMemoryTracking=summary -XX:+UnlockDiagnosticVMOptions -XX:+PrintNMTStatistics -verbose:gc"
 }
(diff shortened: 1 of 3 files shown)


jerboaa (approved):
OK. Do we mention somewhere in our docs that setting `JAVA_DIAGNOSTICS` might set `-XX:+UnlockDiagnosticVMOptions`?

jmtd:
> OK. Do we mention somewhere in our docs that setting JAVA_DIAGNOSTICS might set -XX:+UnlockDiagnosticVMOptions?

No I don't think so, but I'll double-check. [this](https://github.com/jboss-container-images/openjdk/blob/develop/modules/jvm/api/module.yaml#L24) is pretty much all we write about it, I don't think it's separately covered in anything.

The inclusion of UnlockDiagnosticVMOptions is to allow PrintNMTStatistics which was suggested [in this issue back in 2017](https://issues.redhat.com/browse/OPENJDK-111).  Although nobody else has expressed an interest in it since. If there are drawbacks to setting UnlockDiagnosticVMOptions we could omit both.

jerboaa:
My concern is that users might *think* they've never turned that option on, forgetting about non-standard `JAVA_DIAGNOSTICS`, but confusingly (to them) tooling would report it's on. Hence my suggestion to mention it in docs. Perhaps add:

```
Note: This option, if set to true, will set  -XX:+UnlockDiagnosticVMOptions
```

jerboaa (approved):

PR #231, opened 2021-07-23
Revert to rhel-8-* content sets to ensure we trigger rebuilds for CVEs

jmtd:
RPM Errata that ship CVE fixes for RHEL8 (or the UBI8 subset of it) are pushed to CDN repos named `rhel-8-*`. Those pushes are what trigger the automatic rebuild of downstream containers to pick up the fix. Since the fix is not pushed to the `ubi-8-*` repos using this mechanism (some other out-of-band mechanism syncs ubi-8 repos), automatic rebuilds will never trigger for these containers.

I'm discussing this internally (CLOUDBLD-6558), but in the meantime revert to the rhel-8 content sets so we get back automatic rebuilds.

### DIFF
--- a/content_sets_rhel8.yml
+++ b/content_sets_rhel8.yml
@@ -1,12 +1,12 @@
 x86_64:
-- ubi-8-for-x86_64-baseos-rpms__8
-- ubi-8-for-x86_64-appstream-rpms__8
+- rhel-8-for-x86_64-baseos-rpms
+- rhel-8-for-x86_64-appstream-rpms
 ppc64le:
-- ubi-8-for-ppc64le-appstream-rpms__8
-- ubi-8-for-ppc64le-baseos-rpms__8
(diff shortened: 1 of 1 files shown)


jerboaa (commented):
How do we prevent non-UBI 8 content from slipping through the cracks?

jmtd:
> How do we prevent non-UBI 8 content from slipping through the cracks?

The problem is diminished by not installing weak deps (modern cekits default to this), but it's a concern, hence opening CLOUDBLD-6558 to find another solution. I'll also talk to our QE about checks for it.

jmtd:
Hi @jerboaa the situation is worse than I first realised; until we merge this *and ship an image afterwards* container grading is broken too!

jerboaa (approved):
Alright, we need to ship this then and manually check for non-UBI content on release for now.

PR #233, opened 2021-08-10
Add OpenJDK17 container images (test/preview)

jmtd:
Adding this to the develop branch now will mean we will get automatic CI builds (internal at least)  which will help with image development.

The commit which adjusts the jolokia test is a pre-requisite so that we can tag the affected non-jolokia tests for the JDK17 image (which does not carry jolokia)

Thanks!

### DIFF
--- a/modules/jvm/bash/artifacts/opt/jboss/container/java/jvm/java-default-options
+++ b/modules/jvm/bash/artifacts/opt/jboss/container/java/jvm/java-default-options
@@ -135,7 +135,8 @@ gc_config() {
   local maxHeapFreeRatio=${GC_MAX_HEAP_FREE_RATIO:-20}
   local timeRatio=${GC_TIME_RATIO:-4}
   local adaptiveSizePolicyWeight=${GC_ADAPTIVE_SIZE_POLICY_WEIGHT:-90}
-  local gcOptions="${GC_CONTAINER_OPTIONS:--XX:+UseParallelOldGC}"
+  local gcOptions="${GC_CONTAINER_OPTIONS:--XX:+UseParallelGC}"
+
   # for compat reasons we don't set a default value for metaspaceSize
   local metaspaceSize
   # We also don't set a default value for maxMetaspaceSize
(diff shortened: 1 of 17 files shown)


modules/jvm/bash/artifacts/opt/jboss/container/java/jvm/java-default-options
```diff
@@ -135,7 +135,13 @@ gc_config() {
   local maxHeapFreeRatio=${GC_MAX_HEAP_FREE_RATIO:-20}
   local timeRatio=${GC_TIME_RATIO:-4}
   local adaptiveSizePolicyWeight=${GC_ADAPTIVE_SIZE_POLICY_WEIGHT:-90}
-  local gcOptions="${GC_CONTAINER_OPTIONS:--XX:+UseParallelOldGC}"
+
+  if [ "$JAVA_VERSION" -lt 17 ]; then
```
jmtd: I'm actually not sure yet whether we should solve this problem in this way, or introduce a parallel variant of this cekit module.

jerboaa: Why those gymnastics? AFAIK, just using `-XX:+UseParallelGC` will work for JDK 17, 11 and 8 and looking at the enabled flags via `-XX:+Use[...] -XX:+PrintFlagsFinal` for JDK 8 and 11 those are equivalent.

jmtd: Thank you, you're right. I was going from the docs as to the meaning of the flags, rather than the impact it had on the final flags. Re-reading "JEP 366: Deprecate the ParallelScavenge + SerialOld GC Combination" I still feel like the UseParallelOldGC flag is written backwards... but comparing the flag differences between all three JDK versions with the options enabled/disabled is good enough for me. Thanks for the tip on comparing flags like that.

jerboaa (approved):
Seems fine. Perhaps decouple from #234

PR #234, opened 2021-08-11
Tests: de-couple Jolokia testing from other Behave features

jmtd:
This is part of #233, but I'm pushing it separately as it seems to break GitHub CI. I'l get green lights on this PR and then either merge it separately or ignore it in favour of the commit in #233.


jerboaa (approved):
OK.

PR #239, opened 2021-11-02
[OPENJDK-493] Don't install NSS Wrapper for OpenJDK17 containers

jmtd:
Refactor and re-arrange such that the nss_wrapper stuff is not installed into the OpenJDK 17 container.

Testing:

Comparing cekit-generated output source before/after this PR, all changes look OK to me.

Comparing cekit-built binary image output before/after this PR for four images (ubi8/openjdk-{8,11}{,-runtime}) using `container-diff`, the following are the only differences:

 * the runtime images don't have /home/jboss/passwd any more (good)

I will do further checks on the RHEL7 and OpenJDK 17 images, but they require me to jump through a few more hoops to build.

Reviewer note: during the development of this PR I tried a few different approaches. One thing that made life easier was renaming `jboss.container.java.run.bash` to remove the .bash suffix. However that makes the PR "Files changed" view a bit less clear; so I've made sure that's in a parent commit so you can look at the other changes in isolation at the commit level.


jerboaa (approved):
Seems OK to me.

PR #240, opened 2021-11-18
[OPENJDK-528] rework singleton-jdk to use a single RPM transaction

jmtd:
https://issues.redhat.com/browse/OPENJDK-528

The singleton jdk module loops over the output of `rpm -qa` and invokes `rpm -e --nodeps` once per matching line. If a stray JDK is going to be removed, this happens to remove the -devel package before the -headless (normally -headless would go first because -devel depends on it). The can result in some empty directories remaining in /usr/lib/jvm.

Rewrite that loop to call `rpm -e --nodeps` once with all the packages to remove so they're in a single transaction.


jmtd:
I've asked Severin to review this but I think Jiri or Scott should TAL too. Thanks!

jerboaa (approved):
Looks good.

PR #244, opened 2021-11-22
Delete duplicate JAVA_ARGS value in the logs ubi8/openjdk image.

jhuttana:
Could you please review and provide the comments?

### DIFF
--- a/tests/features/java/runtime.feature
+++ b/tests/features/java/runtime.feature
@@ -35,3 +35,15 @@ Feature: Openshift OpenJDK Runtime tests
        | JAVA_DIAGNOSTICS | true                |
     Then container log should contain /deployments/undertow-servlet.jar Hello from CTF test
       And container log should contain -XX:NativeMemoryTracking=summary
+
+  @openjdk
+  @ubi8/openjdk-8
+  @ubi8/openjdk-11
+  @ubi8/openjdk-17
+  @redhat-openjdk-18
+  @openj9
+  Scenario: OPENJDK-474 to ensure JAVA_ARGS is not duplicated in the java command line
+    Given container is started with env
+    | variable  | value  |
+    | JAVA_ARGS | unique |
+  Then container log should not contain unique unique
--- a/modules/run/artifacts/opt/jboss/container/java/run/run-java.sh
+++ b/modules/run/artifacts/opt/jboss/container/java/run/run-java.sh
@@ -237,7 +237,7 @@ startup() {
   else
      args="-jar ${JAVA_APP_JAR}"
   fi
-  log_info "exec $(get_exec_args) java $(get_java_options) -cp \"$(get_classpath)\" ${args} $* ${JAVA_ARGS}"
+  log_info "exec $(get_exec_args) java $(get_java_options) -cp \"$(get_classpath)\" ${args} $*"
   exec $(get_exec_args) java $(get_java_options) -cp "$(get_classpath)" ${args} $*
 }
 


jhuttana:
This fix is to address the Jira issue : https://issues.redhat.com/browse/OPENJDK-474

jmtd:
This looks good. I am just testing it now. I think it would be nice to add a behave test (something in `tests/features/...`) to prove this works. WDYT?

jhuttana:
> This looks good. I am just testing it now. I think it would be nice to add a behave test (something in `tests/features/...`) to prove this works. WDYT?

Yes agree.

jmtd:
Thanks. Please have a go and feel free to ask any questions. I realise running the tests might be awkward as the `cekit test … behave` functionality only works with Docker not Podman, however, if you push a test to a branch in your GitHub fork, GitHub actions *should* run the test  (if it's properly tagged)

jhuttana:
I have added the behave test but I am not sure how to run this single test against given image.
As this is the first time I am trying testing part could you please help?

jhuttana:
@jmtd I have added behave test could you please take a look?

jmtd:
Hi @jhuttana thanks. The test we'll need should check the output of running an image to confirm that JAVA_ARGS is not duplicated. The good news is we can avoid a full s2i build for the test, because the logging line we want to check happens even if no application is present in the image - that is, we can just run the builder image with JAVA_ARGS defined and test the output of the container.  I think we could use  (untested)

```
  Given container is started with env
    | variable  | value  | 
    | JAVA_ARGS | unique |  
  Then container log should not contain unique unique
```

In terms of testing, every time you push to a branch in a PR, GitHub actions is running the test suite on the result. Here's the log output for your existing test, from the last run: https://github.com/jboss-container-images/openjdk/runs/4331417103?check_suite_focus=true#step:8:27331

There's a *lot* for S2I tests. It begins on line 27331 and ends on 28306 near the bottom.

So rather than get the test requirements set up on your host (docker etc) just push to the PR branch and then check the GitHub actions log once it's complete ("Checks" tab, towards the top)

tests/features/java/runtime.feature
```diff
@@ -35,3 +35,13 @@ Feature: Openshift OpenJDK Runtime tests
...
+
+  @openjdk
+  @ubi8/openjdk-8
+  @ubi8/openjdk-11
+  @ubi8/openjdk-17
+  @redhat-openjdk-18
+  @openj9
+  Scenario: Ensure JAVA_ARGS is no longer present in the run script
+    Given s2i build https://github.com/jboss-openshift/openshift-quickstarts from undertow-servlet
+    Then file /opt/jboss/container/java/run/run-java.sh should not contain JAVA_ARGS
```
jmtd: This is a static check for the contents of a file; we need to run a container and check the container log to see what was printed.

tests/features/java/runtime.feature
```diff
@@ -35,3 +35,13 @@ Feature: Openshift OpenJDK Runtime tests
...
       And container log should contain -XX:NativeMemoryTracking=summary
+
+  @openjdk
+  @ubi8/openjdk-8
+  @ubi8/openjdk-11
+  @ubi8/openjdk-17
+  @redhat-openjdk-18
+  @openj9
+  Scenario: Ensure JAVA_ARGS is no longer present in the run script
+    Given s2i build https://github.com/jboss-openshift/openshift-quickstarts from undertow-servlet
```
jmtd: we can avoid a full s2i build for this test

jhuttana:
> Hi @jhuttana thanks. The test we'll need should check the output of running an image to confirm that JAVA_ARGS is not duplicated. The good news is we can avoid a full s2i build for the test, because the logging line we want to check happens even if no application is present in the image - that is, we can just run the builder image with JAVA_ARGS defined and test the output of the container. I think we could use (untested)
> 
> ```
>   Given container is started with env
>     | variable  | value  | 
>     | JAVA_ARGS | unique |  
>   Then container log should not contain unique unique
> ```
> 
> In terms of testing, every time you push to a branch in a PR, GitHub actions is running the test suite on the result. Here's the log output for your existing test, from the last run: https://github.com/jboss-container-images/openjdk/runs/4331417103?check_suite_focus=true#step:8:27331
> 
> There's a _lot_ for S2I tests. It begins on line 27331 and ends on 28306 near the bottom.
> 
> So rather than get the test requirements set up on your host (docker etc) just push to the PR branch and then check the GitHub actions log once it's complete ("Checks" tab, towards the top)

@jmtd I have updated the behave test as you suggested. Could you please take a look?
Still I am not sure how can I do the setup on my host.
I simply push the changes and once the Github actions are complete I will check the details.

jmtd:
Excellent!  looks good: https://github.com/jboss-container-images/openjdk/pull/244/checks#step:8:27341

And the new test fails on the existing images (without this change) which is exactly what we want.

Can you make one last change please and then we're good to merge: Change the Scenario: description to start with OPENJDK-474 (this helps remind us later why we are doing a particular test) and change "Ensure JAVA_ARGS is no longer present in the run script" to describe the specific thing we have changed e.g. "Ensure JAVA_ARGS is not duplicated in the java command line"

Thanks!

jmtd (approved):

PR #247, opened 2021-11-24
Added launcher script for runtime images.

jhuttana:
This fix is to address : https://issues.redhat.com/browse/OPENJDK-413 issue.

### DIFF
--- a/modules/jre/run/artifacts/opt/jboss/container/java/run
+++ b/modules/jre/run/artifacts/opt/jboss/container/java/run
@@ -1,8 +1,19 @@
 #!/bin/bash
+set -e
+
+# Command line arguments given to this script
+args="$*"
+
+source "${JBOSS_CONTAINER_UTIL_LOGGING_MODULE}/logging.sh"
+
+JAVA_OPTS="${JAVA_OPTS:-${JAVA_OPTIONS}}"
+export JAVA_OPTS
+export JAVA_OPTIONS="$JAVA_OPTS"
 
 # Start JVM
 startup() {
   echo "This is the Red Hat UBI8 OpenJDK ${JAVA_VERSION} Runtime container."
+  exec "${JBOSS_CONTAINER_JAVA_RUN_MODULE}/run-java.sh" $args ${JAVA_ARGS}
 }
 
 # =============================================================================
--- a/modules/jre/run/artifacts/opt/jboss/container/java/run-java.sh
+++ b/modules/jre/run/artifacts/opt/jboss/container/java/run-java.sh
@@ -1,0 +1,246 @@
+#!/bin/sh
+
+# Fail on a single failed command
+set -eo pipefail
+
+source "$JBOSS_CONTAINER_UTIL_LOGGING_MODULE/logging.sh"
+
+if [ "${SCRIPT_DEBUG}" = "true" ] ; then
+    set -x
+    log_info "Script debugging is enabled, allowing bash commands and their arguments to be printed as they are executed"
+fi
+
+# ==========================================================
+# Generic run script for running arbitrary Java applications
+#
+# Source and Documentation can be found
+# at https://github.com/fabric8io/run-java-sh
+#
+# ==========================================================
+
+# Error is indicated with a prefix in the return value
+check_error() {
+  local msg=$1
+  if echo ${msg} | grep -q "^ERROR:"; then
+    log_error ${msg}
+    exit 1
+  fi
+}
+
+# Try hard to find a sane default jar-file
+auto_detect_jar_file() {
+  local dir=$1
+
+  # Filter out temporary jars from the shade plugin which start with 'original-'
+  local old_dir=$(pwd)
+  cd ${dir}
+  if [ $? = 0 ]; then
+    local nr_jars=`ls *.jar 2>/dev/null | grep -v '^original-' | wc -l | tr -d '[[:space:]]'`
+    if [ ${nr_jars} = 1 ]; then
+      ls *.jar | grep -v '^original-'
+      exit 0
+    fi
+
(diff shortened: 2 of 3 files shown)


jhuttana:
As per my investigation the run-java.sh script holds good for runtime images also. I dont see anything which we can filter from this launcher script. Already we have filtered s2i related changes like maven, jolokia and prometheus(which in turn has maven related changes for build purpose) from runtime launch scripts. Other than maven(which has major role for application building) rest other content in the launcher script holds good for runtime images also.

jmtd:
If the existing run script is suitable for the runtime images, they should include that module (`jboss.container.java.run`) rather than copy the content to `jboss.container.java.jre.run`. However I am not (yet) convinced :) My concern is it's very easy for us to add features to the images and very, very hard to change or remove them afterwards. Also the intention of hte runtime images was to be as lean as possible so we should really scrutinise anything we add.

The JVM tuning parameters to the existing builder run script are out of date and need reviewing (see e.g. [OPENJDK-78](https://issues.redhat.com/browse/OPENJDK-78) and the related issues). I wouldn't want to copy the existing ones over to the runtime images if we are going to change or remove them anyway.

We definitely do not want `configure_passwd` running in the runtime images.

I think we need to list out all the features of the existing builder run script; the upstream fabric8 run-java script; any that have been specifically requested by e.g. the quarkus team for the runtime images; and compare those lists together, to figure out feature-by-feature what we want. Also, I'll tidy up the OPENJDK JIRAs to make the review work we need to do on the existing tuning parameters clearer, and possibly set those a blockers on this.

jhuttana:
> If the existing run script is suitable for the runtime images, they should include that module (`jboss.container.java.run`) rather than copy the content to `jboss.container.java.jre.run`. However I am not (yet) convinced :) My concern is it's very easy for us to add features to the images and very, very hard to change or remove them afterwards. Also the intention of hte runtime images was to be as lean as possible so we should really scrutinise anything we add.
> 
> The JVM tuning parameters to the existing builder run script are out of date and need reviewing (see e.g. [OPENJDK-78](https://issues.redhat.com/browse/OPENJDK-78) and the related issues). I wouldn't want to copy the existing ones over to the runtime images if we are going to change or remove them anyway.

This helps. I will also take a look.

> 
> We definitely do not want `configure_passwd` running in the runtime images.
> 
> I think we need to list out all the features of the existing builder run script; the upstream fabric8 run-java script; any that have been specifically requested by e.g. the quarkus team for the runtime images; and compare those lists together, to figure out feature-by-feature what we want. Also, I'll tidy up the OPENJDK JIRAs to make the review work we need to do on the existing tuning parameters clearer, and possibly set those a blockers on this.

Setting blockers will be a great help :)

Further,
The exiting run-java.sh script has below functions
```
$ grep -E '^[[:space:]]*([[:alnum:]_]+[[:space:]]*\(\)|function[[:space:]]+[[:alnum:]_]+)' ./modules/run/artifacts/opt/jboss/container/java/run/run-java.sh 
check_error() {
auto_detect_jar_file() {
get_jar_file() {
load_env() {
run_java_options() {
get_java_options() {
format_classpath() {
get_classpath() {
get_exec_args() {
function configure_passwd() {
startup() {
```

Except configure_passwd() , to me it looks like it covers all the features requested by quarkus team here: https://issues.redhat.com/browse/OPENJDK-413
Even some JVM tuning parameters are expected. In that case we can try to customize the JVM parameters specifically for runtime images.

jhuttana:
I was going through https://issues.redhat.com/browse/OPENJDK-78 and in the issue description found  point which :
"we should just start without JVM parameters and let us add some if we need, rather than trying to hack around to remove parameters."
This is in fact is good thing to let user decide which JVM parameters are suitable for their application.  WDYT?

jmtd:
Hi @jhuttana, I've thought about the test a bit more and we don't need a new behave step: I was thinking that the test needed to be run as a different user to default, but all we need to do is check the output of `stat` on /home/jboss. Would you mind please trying to write a test that does that? Take a look at the  existing tests to find one that runs a specific command in a container and checks the container log for something. Correct output is like this:

```
  File: /home/jboss
Device: 3fh/63d	Inode: 1148428     Links: 2
Access: (0771/drwxrwx--x)  Uid: (  185/   jboss)   Gid: (    0/    root)
Access: 2022-01-06 13:25:01.459525187 +0000
Modify: 2022-01-06 13:24:29.000000000 +0000
Change: 2022-01-06 13:25:01.451525177 +0000
 Birth: 2022-01-06 13:25:01.331525026 +0000
```

jmtd:
>  This is in fact is good thing to let user decide which JVM parameters are suitable for their application. WDYT?

The challenge is reaching that without breaking backwards compatibility for existing customers.

jhuttana:
Ok.

maxandersen:
> The challenge is reaching that without breaking backwards compatibility for existing customers.

maybe I'm missing something here but haven't backwards compatibility not already been broken by not having run-java.sh available ? :)

jmtd:
> maybe I'm missing something here but haven't backwards compatibility not already been broken by not having run-java.sh available ? :)

These images have never shipped with a runtime script, so no. My comment RE backwards compatibility was in reply to the suggestion that we drop all the JVM tuning parameters by default for the existing run script - which is in the builder images.

gsmet:
@jhuttana @jmtd any chance some progress could be made on this soon? We are eagerly waiting for this for the Quarkus containers.

Thanks!

jmtd:
@gsmet we have made some forward progress on this today. Sadly the last couple of weeks were interrupted by two CVEs and an OpenJDK regression.

gsmet:
Cool, thanks for letting me know!

jmtd:
This was superseded by #277.

PR #250, opened 2021-11-26
OPENJDK-533: test $HOME/passwd can be written to

jmtd:
run-java.sh has a line which attempts to write to $HOME/passwd, which is
part of the solution for backwards compatibility on OCP 3.11
(OPENJDK-312). This was made writeable by a common module but that was
moved to a new nss-wrapper module in OPENJDK-493, which is not installed
in OpenJDK 17 containers.

Two problems result for the OpenJDK-17 image:

1. Starting a downstream (S2I) container with a random UID (OpenShift
   default) terminates when the attempt to write to /home/jboss/passwd
   fails
2. for some container runtimes (at least Docker), the container will
   refuse to start if the running uid can't chdir to /home/jboss

That change also moved changing the /home/jboss permissions. Partially
revert those changes and check to ensure the file can be written before
attempting to.

----

Testing performed:

 • local s2i builds, verified the /home/jboss +x change resolves the docker issue
 • `oc new-app` (OCP 4.x cluster) with a quarkus quickstart, pod runs

I'm surprised that the Behave tests didn't catch these issues, and I'm looking now
to find out why.

### DIFF
--- a/modules/run/artifacts/opt/jboss/container/java/run/run-java.sh
+++ b/modules/run/artifacts/opt/jboss/container/java/run/run-java.sh
@@ -220,7 +220,11 @@ get_exec_args() {
 # Ensure that the running UID has the "jboss" passwd metadata
 # XXX: Maybe we should make this an entrypoint for the image?
 function configure_passwd() {
-  sed "/^jboss/s/[^:]*/$(id -u)/3" /etc/passwd > "$HOME/passwd"
+  # OPENJDK-533: this file is only writeable if the image uses the
+  # nss_wrapper module. ubi8/openjdk-17 does not.
+  if [ -w "$HOME/passwd" ]; then
+    sed "/^jboss/s/[^:]*/$(id -u)/3" /etc/passwd > "$HOME/passwd"
+  fi
 }
 
 # Start JVM
--- a/modules/user/configure.sh
+++ b/modules/user/configure.sh
@@ -6,3 +6,7 @@ set -e
 # This ID is registered static ID for the JBoss EAP product
 # on RHEL which makes it safe to use.
 groupadd -r jboss -g 185 && useradd -u 185 -r -g root -G jboss -m -d /home/jboss -s /sbin/nologin -c "JBoss user" jboss
+
+# OPENJDK-533: Some container runtimes (Docker) will fail to start if
+# the running UID cannot chdir to $HOME
+chmod og+x /home/jboss


modules/run/artifacts/opt/jboss/container/java/run/run-java.sh
```diff
@@ -220,7 +220,9 @@ get_exec_args() {
 # Ensure that the running UID has the "jboss" passwd metadata
 # XXX: Maybe we should make this an entrypoint for the image?
 function configure_passwd() {
-  sed "/^jboss/s/[^:]*/$(id -u)/3" /etc/passwd > "$HOME/passwd"
+  if [ -w "$HOME/passwd" ]; then
```
jerboaa, line 225: Please add a comment as to why that check is needed. It's easy to forget why it was added otherwise.

jerboaa (commented):
@jmtd Is there a reason why https://github.com/jboss-container-images/openjdk/blob/develop/tests/features/java/openjdk_s2i.feature isn't tagged with `@ubi8/openjdk-17`. It might be that running that test would have caught this issue? Or did it run and it passed?

jmtd:
@jerboaa <s>it's tagged `\@openjdk` which should mean it's run for any image prefixed with `openjdk/` including the 17 images</s> ­-- ah no, `@openjdk` will only catch `openjdk/openjdk-11-rhel7`. This should have the tag `@ubi8` added.

None of the tests it seems would catch bug 1) above, since they almost all override the  CMD to do somehing like stat or grep, so run-java.sh was not getting exercised. I've been working on addressing that, but there are some complexities trying to do an s2i test and a non-default uid test at the same time.

For issue 2), I would have thought nearly all the tests should fail, so I cannot explain yet why they didn't.

modules/run/artifacts/opt/jboss/container/java/run/run-java.sh
jmtd: Good idea.

jerboaa (commented):
I'd really like to have a regression test for this along the lines of:

```
$ podman run --rm -ti -u 12324 registry.access.redhat.com/ubi8/openjdk-17
Starting the Java application using /opt/jboss/container/java/run/run-java.sh ...
ERROR Neither $JAVA_MAIN_CLASS nor $JAVA_APP_JAR is set and 0 JARs found in /deployments (1 expected)
/opt/jboss/container/java/run/run-java.sh: line 45: cd: /home/jboss: Permission denied
```

The test could assert that it should not contain `Permission denied`.

It would just too easy to regress again otherwise.

sefroberg:
> 

I can add this into our regression suite. Is the rule, we should be able to run the container whichever user we pass in? In your example you are passing in "12324"

jerboaa:
> > 
> 
> I can add this into our regression suite. Is the rule, we should be able to run the container whichever user we pass in? In your example you are passing in "12324"

OK. Yes, in the OpenShift case containers run with random user ids (what I mimic with `-u 12324`). So basic tests would need to verify that images aren't broken when run that way.

jmtd:
> I'd really like to have a regression test for this along the lines of:

I've got a behave test like this and a few other related WIP test changes in a branch (OPENJDK-533-passwd-perms-wip) which I'll be working on fixing up over the next couple of days whilst I'm at a conference.

jmtd:
Merging this in the current state as we've had to ship it in a hurry. I'll raise a new issue to reflect we need to improve the behave testing to verify this issue.

PR #251, opened 2021-11-26
Updated container metadata for documentation to point at new docs.

jhuttana:


jmtd:
Thanks - for the builder images that's perfect. For the runtime images, it's not appropriate (documents s2i features they don't have). However, the old value is also problematic! Can we delete the "usage" label from the runtime images for now?

jhuttana:
> Thanks - for the builder images that's perfect. For the runtime images, it's not appropriate (documents s2i features they don't have). However, the old value is also problematic! Can we delete the "usage" label from the runtime images for now?

Sure will do that

jhuttana:
Deleted in-appropriate metadata from runtime descriptors.

jmtd (approved):
👍🏼

PR #252, opened 2021-11-26
Added LANG environment variable to ubi8-based images.

jhuttana:

### DIFF
--- a/tests/features/java/openjdk.feature
+++ b/tests/features/java/openjdk.feature
@@ -76,3 +76,15 @@ Feature: Miscellaneous OpenJDK-related unit tests
     | command | ls -1 /usr/lib/jvm |
     Then available container log should not contain java-1.8.0
     Then available container log should not contain java-11
+
+  @ubi8/openjdk-8
+  @ubi8/openjdk-8-runtime
+  @ubi8/openjdk-11
+  @ubi8/openjdk-11-runtime
+  @ubi8/openjdk-17
+  @ubi8/openjdk-17-runtime
+  Scenario: Ensure LANG is defined and contains UTF-8
+    When container is started with args
+    | arg     | value                                  |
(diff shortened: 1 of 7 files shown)


jhuttana:
```
$ podman run --rm -ti localhost/ubi8/openjdk-8-runtime bash
[jboss@4705f0666522 ~]$ echo $LANG
LANG=C.utf8
```

jerboaa (commented):
Please add a behave test which asserts that `java -XshowSettings:properties -version 2>&1 | grep file.encoding` has value `file.encoding = UTF-8` when run with those images. Add them for example here: https://github.com/jboss-container-images/openjdk/blob/develop/tests/features/java/openjdk.feature. Adding a test for the LANG variable would be useful too.

jhuttana:
> Please add a behave test which asserts that `java -XshowSettings:properties -version 2>&1 | grep file.encoding` has value `file.encoding = UTF-8` when run with those images. Add them for example here: https://github.com/jboss-container-images/openjdk/blob/develop/tests/features/java/openjdk.feature. Adding a test for the LANG variable would be useful too.


Added behave test and manually tested as below:

Ex:
```
$ podman run --rm -ti localhost/ubi8/openjdk-17-runtime bash -c "java -XshowSettings:properties -version 2>&1 | grep "file.encoding""
 file.encoding = UTF-8
```

jhuttana:
The special character pipe "|"  in behave test action table was throwing "Malformed Table" error.
So I just used `bash -c "$JAVA_HOME/bin/java -XshowSettings:properties -version"` instead of `bash -c "$JAVA_HOME/bin/java -XshowSettings:properties -version 2>&1 | grep file.encoding"`

jmtd (approved):
LGTM!

PR #259, opened 2022-01-04
[OPENJDK-556] Moved chmod instructions for $HOME (/home/jboss) from nss-wrapper mod…

jhuttana:

### DIFF
--- a/tests/features/general.feature
+++ b/tests/features/general.feature
@@ -1,0 +1,10 @@
+Feature: Miscellaneous general settings unit tests
+
+  @ubi8
+  @redhat-openjdk-18/openjdk18-openshift
+  @openjdk/openjdk-11-rhel7
+  Scenario: Check the attributes of /home/jboss using stat
+    When container is started with args
+    | arg     | value            |
+    | command | stat /home/jboss |
+    Then available container log should contain Access: (0771/drwxrwx--x)
--- a/modules/user/configure.sh
+++ b/modules/user/configure.sh
@@ -6,6 +6,7 @@ set -e
 # This ID is registered static ID for the JBoss EAP product
 # on RHEL which makes it safe to use.
 groupadd -r jboss -g 185 && useradd -u 185 -r -g root -G jboss -m -d /home/jboss -s /sbin/nologin -c "JBoss user" jboss
+chmod ug+rwX /home/jboss
 
 # OPENJDK-533: Some container runtimes (Docker) will fail to start if
 # the running UID cannot chdir to $HOME
--- a/modules/util/nss-wrapper/configure.sh
+++ b/modules/util/nss-wrapper/configure.sh
@@ -3,4 +3,4 @@ set -euo pipefail
 
 # set up a copy of the passwd file which nss_wrapper will use.
 cp /etc/passwd /home/jboss/passwd
-chmod ug+rwX /home/jboss /home/jboss/passwd
+chmod ug+rwX /home/jboss/passwd


jhuttana:
I feel we can delete modules/util/nss-wrapper/configure.sh as we have moved its contents to module/user/configure.sh.
That needs changes to modules/util/nss-wrapper/module.yaml file. 
Could you please confirm whether this makes sense?

jmtd:
> I feel we can delete modules/util/nss-wrapper/configure.sh as we have moved its contents to module/user/configure.sh.

We only want to move the `chmod` for the home directory to the user module: we don't want to move or copy the instructions relating to the passwd file. And we don't want to add the nss wrapper stuff to the openjdk 17 image: it's only in the 8 and 11 images for backwards compatibility.

jhuttana:
> > I feel we can delete modules/util/nss-wrapper/configure.sh as we have moved its contents to module/user/configure.sh.


> We only want to move the `chmod` for the home directory to the user module: we don't want to move or copy the instructions relating to the passwd file. 

Okay sure.

> And we don't want to add the nss wrapper stuff to the openjdk 17 image: it's only in the 8 and 11 images for backwards compatibility.

Here: https://issues.redhat.com/browse/OPENJDK-556 , you mentioned  that JDK17 builder also needs this change. So I thought this change is necessary in its descriptor.

jhuttana:
@jmtd I have addressed the review comment. Could you please take a look?

jmtd:
That looks right, thanks: I'll try some builds and see what the results look like.

jhuttana:
> That looks right, thanks: I'll try some builds and see what the results look like.


Sure :) Thank you!

jmtd:
Thank you the change looks good. I've compared 11, 17 and 17-runtime before and after.

It would be good to have a test, although I'm not sure that the existing Behave test steps are sufficient. What command would you run to check this? I wonder if this test step would be enough: https://github.com/cekit/behave-test-steps/blob/924a8ee7cbc3f4b5e0b9339843a51ff11bb1ac5f/steps/container_steps.py#L147

jhuttana:
> Thank you the change looks good. I've compared 11, 17 and 17-runtime before and after.
> 
> It would be good to have a test, although I'm not sure that the existing Behave test steps are sufficient.

ok. Even I am not sure ;)

> What command would you run to check this? 

To verify this change I used
```
$ podman run --rm -ti localhost/ubi8/openjdk-17:latest ls -l /home/
total 4
drwxrwx--x. 3 jboss root 4096 Jan  5 11:52 jboss
```
before and after the change

>I wonder if this test step would be enough: https://github.com/cekit/behave-test-steps/blob/924a8ee7cbc3f4b5e0b9339843a51ff11bb1ac5f/steps/container_steps.py#L147

Not sure :) This is new to me :)

jmtd:
Hi @jhuttana, I've thought about the test a bit more and we don't need a new behave step: I was thinking that the test needed to be run as a different user to default, but all we need to do is check the output of `stat` on /home/jboss. Would you mind please trying to write a test that does that? Take a look at the  existing tests to find one that runs a specific command in a container and checks the container log for something. Correct output is like this:

```
  File: /home/jboss
Device: 3fh/63d	Inode: 1148428     Links: 2
Access: (0771/drwxrwx--x)  Uid: (  185/   jboss)   Gid: (    0/    root)
Access: 2022-01-06 13:25:01.459525187 +0000
Modify: 2022-01-06 13:24:29.000000000 +0000
Change: 2022-01-06 13:25:01.451525177 +0000
 Birth: 2022-01-06 13:25:01.331525026 +0000
```

jhuttana:
Sure. I will give a try

jhuttana:
Here I see that commands are run to find something in the log as a part of test execution.
https://github.com/jboss-container-images/openjdk/blob/develop/tests/features/java/openjdk.feature

I will add something similiar test and create a new file tests/features/general.feature

tests/features/general.feature
```diff
@@ -0,0 +1,13 @@
...
+  @ubi8/openjdk-8-runtime
+  @ubi8/openjdk-11
+  @ubi8/openjdk-11-runtime
+  @ubi8/openjdk-17
+  @ubi8/openjdk-17-runtime
+  Scenario: Check the attributes of /home/jboss using stat
+    When container is started with args
+    | arg     | value            |
+    | command | stat /home/jboss |
+    Then available container log should contain File:
```
jmtd: Replace `File:` with one of the lines of output we would expect to see from `stat`, which would prove the permissions are correct, e.g. `Access: (0771/drwxrwx--x)  Uid: (  185/   jboss)   Gid: (    0/    root)`

tests/features/general.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Miscellaneous general settings unit tests
+
+  @ubi8/openjdk-8
+  @ubi8/openjdk-8-runtime
+  @ubi8/openjdk-11
+  @ubi8/openjdk-11-runtime
+  @ubi8/openjdk-17
+  @ubi8/openjdk-17-runtime
```
jmtd: these @-prefixed lines select which images the tests are run against. Here we specify only the ubi8-based ones. I think that this test should run against all the images. The way the tag matching works, if you specify `@ubi8`, then any image beginning wtih  `@ubi8` (i.e. all of these ones) would match. So these lines could be replaced with just `@ubi8`. But please also add lines `@redhat-openjdk-18/openjdk18-openshift` and `@openjdk/openjdk-11-rhel7` to match our other two images.

jmtd (commented):
This is good, thanks! I've left a couple of comments on the test, after those are addressed it's good to go.

jhuttana:
I have addressed the comment and checks passed

jmtd:
Great! test looks good. Here's an example of it run: https://github.com/jboss-container-images/openjdk/runs/4763240877?check_suite_focus=true#step:8:137

PR #271, opened 2022-03-23
Remove OpenJ9 images from ImageStream metadata

jmtd:
Syntax tested with jq and semantic tested with "oc create".

One open question I have before merging: What happen if someone has an
OpenShift instance with running pods based on these images, and then they
import our updated ImageStream metadata that deletes those images?


jmtd:
already done

PR #274, opened 2022-03-28
Remove OpenJ9 images.

jmtd:
I think these are no longer necessary. @levivic can you confirm?


levivic:
hi @jmtd , as John mentioned in his comment https://issues.redhat.com/browse/RHMWRT-692?focusedCommentId=19986124&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-19986124, looks like we will still look for other options. Maybe we should keep it as it is for now, just in case we may need it again in the future?

jmtd:
hi @levivic , it would not be hard for us to revert this change later on if we did return to building them, but in the meantime it would make openjdk development easier to remove the unrelated files. Had we not had the change of direction, I was planning to talk to you about moving the j9 sources to a fork repo anyway.

jmtd:
Merging as described in my last comment: this can always be reversed. Meanwhile, it's simplifying the code base for other developments

PR #275, opened 2022-03-28
[OPENJDK-628] run: explicitly specify bash; simplify shell handling

jmtd:
There was some code to work around non-bash shells that is not
necessary if we are using bash at all times. Clean it up and
explicitly require bash in the shebang line.

https://issues.redhat.com/projects/OPENJDK/issues/OPENJDK-628


jerboaa (approved):
LGTM

PR #277, opened 2022-04-06
[OPENJDK-413] Revert "Changes to add CMD for runtime images"

jhuttana:
https://issues.redhat.com/browse/OPENJDK-413

----

Could you please review the changes?
Image sizes:
Before:
localhost/ubi8/openjdk-17-runtime                              latest      ff254a722004  6 days ago     426 MB
localhost/ubi8/openjdk-17-runtime                              1.11        ff254a722004  6 days ago     426 MB
localhost/ubi8/openjdk-11-runtime                              1.11        f90246ad32e5  6 days ago     425 MB
localhost/ubi8/openjdk-11-runtime                              latest      f90246ad32e5  6 days ago     425 MB
localhost/ubi8/openjdk-8-runtime                               latest      44dedd96fea7  6 days ago     370 MB
localhost/ubi8/openjdk-8-runtime                               1.11        44dedd96fea7  6 days ago     370 MB

After:
localhost/ubi8/openjdk-8-runtime                               1.11        5f2c4c48395f  5 minutes ago  595 MB
localhost/ubi8/openjdk-8-runtime                               latest      5f2c4c48395f  5 minutes ago  595 MB
localhost/ubi8/openjdk-11-runtime                              1.11        f74f2bfd2249  6 minutes ago  648 MB
localhost/ubi8/openjdk-11-runtime                              latest      f74f2bfd2249  6 minutes ago  648 MB
localhost/ubi8/openjdk-17-runtime                              latest      2d60e4b70f9d  8 minutes ago  450 MB
localhost/ubi8/openjdk-17-runtime                              1.11        2d60e4b70f9d  8 minutes ago  450 MB

PR #278, opened 2022-04-06
Revert "Changes to add CMD for runtime images"

jhuttana:


jmtd:
I don't think we should merge this as-is, since it will cause the runtime images to be less feature ful lthan they are now. The same changes are part of #277 , so we should just merge them as part of that change when it is ready.

PR #285, opened 2022-05-04
[OPENJDK-559] Set MaxRAMPercentage=80 by default

jmtd:
<https://issues.redhat.com/browse/OPENJDK-559>

Permit overriding this default with MAX_RAM_PERCENTAGE.


modules/jvm/api/module.yaml
```diff
@@ -21,6 +21,9 @@ envs:
 - name: JAVA_MAX_INITIAL_MEM
   description: Is used when no `-Xms` option is given in **JAVA_OPTS**. This is used to calculate the maximum value of the initial heap memory. If used in a container without any memory constraints for the container then this option has no effect. If there is a memory constraint then `-Xms` is limited to the value set here. The default is 4096MB which means the calculated value of `-Xms` never will be greater than 4096MB. The value of this variable is expressed in MB.
   example: "4096"
+- name: MAX_RAM_PERCENTAGE
+  description: Set this to alter the value for the `-XX:MaxRAMPercentage` parameter from the default value of 80, corresponding to 80%.
+  example: "100"
```
jerboaa: An example of `100` seems dubious. It suggests setting the heap to `100%` makes sense leaving no room for non-heap memory like thread stacks, code cache, metaspace, etc. Perhaps `example: 90`?

modules/jvm/bash/artifacts/opt/jboss/container/java/jvm/java-default-options
```diff
@@ -160,4 +166,4 @@ error_handling() {
 initialize_container_limits > /dev/null
 
 ## Echo options, trimming trailing and multiple spaces
-echo "$(initial_memory) $(max_memory) $(gc_config) $(diagnostics) $(error_handling)" | awk '$1=$1'
+echo "$(initial_memory) $(max_memory) $(max_ram_percentage) $(gc_config) $(diagnostics) $(error_handling)" | awk '$1=$1'
```
jerboaa: What are the defaults for `initial_memory` and `max_memory`? Would those set `-Xms` and `-Xmx` values by default (i.e. if nothing has been specified by the user? If so, setting `-XX:MaxRAMPercentage` doesn't make sense as explicit `-Xmx` settings take precedence.

tests/features/java/openjdk.feature
```diff
@@ -85,3 +85,19 @@ Feature: Miscellaneous OpenJDK-related unit tests
     | arg     | value                                  |
     | command | bash -c "$JAVA_HOME/bin/java -XshowSettings:properties -version" |
     Then available container log should contain file.encoding = UTF-8
+
+  @redhat-openjdk-18/openjdk18-openshift
+  @openjdk/openjdk-11-rhel7
+  @ubi8
+  Scenario: Ensure default MaxRAMPercentage is specified as 80
+    When container is ready
+    Then available container log should contain -XX:MaxRAMPercentage=80
```
jerboaa: The test should verify that no `-Xmx` options are present as explained above.

jerboaa (commented):
Looks mostly good. A few questions...

modules/jvm/bash/artifacts/opt/jboss/container/java/jvm/java-default-options
jmtd: Ugh. I haven't been able to quickly confirm what circumstances will result in -Xmx and -Xms being set by the launch scripts. My position is: if the user effectively specifies those directly, then we should not detect/prevent that, just as they could do so by running `java` themselves: let the runtime decide what to do. But yes, if there's an indirect way we cause `-Xmx` or `-Xms` to be specified we should ensure that this makes sense wrt to this new option.

I think the circumstance where that might happen is you have a container runtime imposed memory limit on a cgroups v1 system. I need to make some adjustments to my setup to test that scenario properly yet.

(I am looking forward to dumping all this stuff in 9)

jmtd:
Note to self: got `Improperly specified VM option 'MaxRAMPercentage=80'` whilst trying to test a cgroups v1 scenario, dig deeper.

jerboaa:
> Note to self: got `Improperly specified VM option 'MaxRAMPercentage=80'` whilst trying to test a cgroups v1 scenario, dig deeper.

Seems JDK 8 specific. Use `80.0` over `80`. The double notation seems to work for later JDKs too.

jmtd:
We can re-use the exising  `JAVA_MAX_MEM_RATIO` and `JAVA_INITIAL_MEM_RATIO` container parameters, which were always defined as a percentage of available memory, to map straight through to `-XX:MaxRAMPercentage` (etc.), and remove any/all of the cgroups-checking that the current scripts try to do.

The sticking point is the existing `JAVA_MAX_INITIAL_MEM` parameter, which is defined as an absolute value (not a percentage) and used as a ceiling value for `JAVA_INITIAL_MEM_RATIO`. I think we should at least deprecate it, and ideally remove it altogether, but if we can't do that immediately for backwards-compat reasons, we have to think about how to handle it. Perhaps, I'm hoping, we could blindly map it to `-Xms`, I need to check what the JVM does in that situation. I think `-Xms` will override `-XX:InitialRAMPercentage` so things might just work.

modules/jvm/api/module.yaml
```diff
@@ -21,6 +21,9 @@ envs:
 - name: JAVA_MAX_INITIAL_MEM
   description: Is used when no `-Xms` option is given in **JAVA_OPTS**. This is used to calculate the maximum value of the initial heap memory. If used in a container without any memory constraints for the container then this option has no effect. If there is a memory constraint then `-Xms` is limited to the value set here. The default is 4096MB which means the calculated value of `-Xms` never will be greater than 4096MB. The value of this variable is expressed in MB.
   example: "4096"
+- name: MAX_RAM_PERCENTAGE
+  description: Set this to alter the value for the `-XX:MaxRAMPercentage` parameter from the default value of 80, corresponding to 80%.
+  example: "90"
```
jerboaa: I think this should be `90.0`

modules/jvm/bash/artifacts/opt/jboss/container/java/jvm/java-default-options
```diff
@@ -76,6 +76,12 @@ max_memory() {
   fi
 }
 
+# Adjust the -XX:MaxRAMPercentage value
+max_ram_percentage() {
+  # JVM default is currently 25; we default to 80
+  echo "-XX:MaxRAMPercentage=${MAX_RAM_PERCENTAGE:-80}"
```
jerboaa: `80.0` over `80` for JDK 8 compatibility.

jmtd:
@jerboaa this is now ready for review :)

jerboaa (approved):
LGTM.

PR #287, opened 2022-05-11
[OPENJDK-776] WIP: UBI9 images

jmtd:
This is a work-in-progress but raising the PR now since time is ticking
so SME review can begin if possible.

The starting point is to copy the UBI8 images exactly.

The next step is to slim the UBI9 ones down module by module.

The general approach is to: identify each module to slim; make a copy
with a higher version number (generally 9.0); adjust the *older images*
so they pull in the lower version number; hack away at the newer
version.

The rationale behind modifying the older image descriptors rather than
the new ones, despite that meaning some source-level changes to the
older images, is to ensure that the *default happy path* is the newer
modules, and the legacy/deprecated ones are treated as exceptional.

Doing it this way does result in the precise module order for older
images changing. I have tried to do a binary image comparison after
each such change to ensure that there are no significant differences.
Insignificant differences include things like RPM package DB and logs
not being identical due to a different RPM install order (etc)

https://issues.redhat.com/browse/OPENJDK-776


jmtd:
I'm open to other approaches to managing the source than the approach I've used here. In particular I wonder now whether to keep different RHEL major versions in distinct git branches. There are pros and cons to each approach. Considering that and any switch should probably be done at a different time.

jmtd:
GitHub CI is not going to be able to build or test RHEL9 images prior to GA. I can build and run test suites and circulate results internally.

modules/jvm/api/v9/module.yaml
```diff
@@ -0,0 +1,59 @@
...
+- name: JAVA_MAX_MEM_RATIO
+  description: Is used when no `-Xmx` option is given in **JAVA_OPTS**. This is used to calculate a default maximal heap memory based on a containers restriction. If used in a container without any memory constraints for the container then this option has no effect. If there is a memory constraint then `-Xmx` is set to a ratio of the container available memory as set here. The default is `50` which means 50% of the available memory is used as an upper boundary. You can skip this mechanism by setting this value to `0` in which case no `-Xmx` option is added.
+- name: JAVA_INITIAL_MEM_RATIO
+  description: Is used when no `-Xms` option is given in **JAVA_OPTS**. This is used to calculate a default initial heap memory based on the maximum heap memory. If used in a container without any memory constraints for the container then this option has no effect. If there is a memory constraint then `-Xms` is set to a ratio of the `-Xmx` memory as set here. The default is `25` which means 25% of the `-Xmx` is used as the initial heap size. You can skip this mechanism by setting this value to `0` in which case no `-Xms` option is added.
+  example: "25"
```
jerboaa, line 20: Should we drop those in favor of `-XX:MaxRAMPercentage` and `-XX:InitialRAMPercentage` (and relevant related env var)?

modules/jvm/api/v9/module.yaml
```diff
@@ -0,0 +1,59 @@
...
+  description: Set this to get some diagnostics information to standard output when things are happening. **Note:** This option, if set to true, will set `-XX:+UnlockDiagnosticVMOptions`. **Disabled by default.**
+  example: "true"
+- name: JAVA_DEBUG
+  description: If set remote debugging will be switched on. **Disabled by default.**
+  example: "true"
+- name: JAVA_DEBUG_PORT
+  description: Port used for remote debugging. Defaults to *5005*.
+  example: "8787"
+- name: CONTAINER_CORE_LIMIT
+  description: A calculated core limit as described in https://www.kernel.org/doc/Documentation/scheduler/sched-bwc.txt.
```
jerboaa, line 34: Drop this setting in light of cgroups v1 and v2 support?

modules/jvm/api/v9/module.yaml
```diff
@@ -0,0 +1,59 @@
...
+  example: "2"
+- name: CONTAINER_MAX_MEMORY
```
jerboaa, line 36: Not sure what this does. Is this setting `-XX:MaxRAM`?

modules/jvm/v9/artifacts/opt/jboss/container/java/jvm/container-limits
```diff
@@ -0,0 +1,35 @@
...
+}
+
+min() {
+  printf "%s\n" "$@" | sort -g | head -n1
+}
+
+local max_mem="$(container_memory)"
+if [ x$max_mem != x ]; then
+  export CONTAINER_MAX_MEMORY="$max_mem"
+fi
```
jerboaa, line 35: This looks like old stuff which might be inferior to what the JVM is doing.

modules/jvm/v9/artifacts/opt/jboss/container/java/jvm/java-default-options
```diff
@@ -0,0 +1,146 @@
...
+
+  # Check for the 'real memory size' and calculate mx from a ratio
+  # given (default is 50%)
+  if [ "x$CONTAINER_MAX_MEMORY" != x ]; then
+    local max_mem="${CONTAINER_MAX_MEMORY}"
+    local ratio=${JAVA_MAX_MEM_RATIO:-${__DEFAULT_JAVA_MAX_MEM_RATIO}}
+    local mx=$(echo "${max_mem} ${ratio} 1048576" | awk '{printf "%d\n" , ($1*$2)/(100*$3) + 0.5}')
+    echo "-Xmx${mx}m"
+  fi
+}
```
jerboaa, line 68: This should get replaced by setting `-XX:MaxRAMPercentage=80` or if we really need 1-to-1 compat to `50`.

modules/jvm/v9/artifacts/opt/jboss/container/java/jvm/java-default-options
```diff
@@ -0,0 +1,146 @@
...
+    local initial_ratio=${JAVA_INITIAL_MEM_RATIO:-${__DEFAULT_JAVA_INITIAL_MEM_RATIO}}
+    local ms=$(echo "${max_mem} ${max_ratio} ${initial_ratio} 1048576" | awk '{printf "%d\n" , ($1*(($2*$3)/10000))/$4 + 0.5}')
+    local max_initial_memory=${JAVA_MAX_INITIAL_MEM:-${__DEFAULT_JAVA_MAX_INITIAL_MEM}}
+    if [ "${ms}" -lt "${max_initial_memory}" ] ; then
+      echo "-Xms${ms}m"
+    else
+      echo "-Xms${max_initial_memory}m"
+    fi
+  fi
+}
```
jerboaa, line 91: Same as for `max_memory`, but use `-XX:InitialRAMPercentage=XX` instead.

modules/run/v9/artifacts/opt/jboss/container/java/run/run-java.sh
```diff
@@ -0,0 +1,228 @@
...
+
+  local args
+  cd ${JAVA_APP_DIR}
+  if [ "x${JAVA_MAIN_CLASS}" != x ] ; then
+     args="${JAVA_MAIN_CLASS}"
+  else
+     args="-jar ${JAVA_APP_JAR}"
+  fi
+  log_info "exec $(get_exec_args) java $(get_java_options) -cp \"$(get_classpath)\" ${args} $*"
+  exec $(get_exec_args) java $(get_java_options) -cp "$(get_classpath)" ${args} $*
```
jerboaa, line 223: Consider an option which execs `java SOME_OPTIONS -cp  ...` as an alternative to `java $(get_java_options)...` Where `SOME_OPTIONS` is taken verbatim from user settings.

modules/s2i/core/bash/v9/artifacts/opt/jboss/container/s2i/core/s2i-core
```diff
@@ -0,0 +1,151 @@
...
+  if [ -d "${1}/${S2I_SOURCE_DATA_DIR}" ]; then
+    if [ -z "${S2I_TARGET_DATA_DIR}" ]; then
+      log_warning "Unable to copy data files.  No target directory specified for S2I_TARGET_DATA_DIR"
+    else
+      if [ ! -d "${S2I_TARGET_DATA_DIR}" ]; then
+        log_info "S2I_TARGET_DATA_DIR does not exist, creating ${S2I_TARGET_DATA_DIR}"
+        mkdir -pm 775 "${S2I_TARGET_DATA_DIR}"
+      fi
+      log_info "Copying app data from $(realpath --relative-to ${S2I_SOURCE_DIR} ${1}/${S2I_SOURCE_DATA_DIR}) to ${S2I_TARGET_DATA_DIR}..."
+      rsync -rl --out-format='%n' "${1}/${S2I_SOURCE_DATA_DIR}"/ "${S2I_TARGET_DATA_DIR}"
```
jerboaa, line 73: In passing, perhaps reconsider whether rsync is needed in the OpenJDK container images. It's not clear what the use-case for this is, though.

redhat/ubi9-openjdk-17-runtime.yaml
```diff
@@ -0,0 +1,9 @@
+osbs:
+  configuration:
+    container:
+      platforms:
+        only:
+        - x86_64
```
jerboaa, line 6: Are runtime images a `x86_64` only feature? if so, why?

ubi9-openjdk-17-runtime.yaml
```diff
@@ -0,0 +1,51 @@
+# This is an Image descriptor for Cekit
+
+schema_version: 1
+
+from: "ubi9-minimal"
+name: &name "ubi9/openjdk-17-runtime"
+version: &version "1.13"
+description: "Image for Red Hat OpenShift providing OpenJDK 17 runtime"
```
jerboaa, line 8: Shouldn't this have a more generic explanation? Like, for example, `UBI 9 OpenJDK 17 runtime image`

jerboaa (commented):
First pass. Take with a grain of salt. It's a massive patch ;-)

modules/jvm/api/v9/module.yaml
jmtd: It's not intended to be set by the user, it's set by the scripts which inspect the cgroups *v1* filesystem layouts for memory limits. It influences the setting of -Xmx I think, and possibly others. It should probably go.

modules/s2i/core/bash/v9/artifacts/opt/jboss/container/s2i/core/s2i-core
jmtd: I'm fairly sure we can get rid of it. Whoever introduced it (not me ;)) from dim and distant memory may have been working around some issues with `cp -a` or similar which might not apply any more.

redhat/ubi9-openjdk-17-runtime.yaml
jmtd: No, but at present the internal Compose URIs I have access to for building RHEL9 images are only populated for X86_64, so without it, OSBS attempts multiarch and fails. I am actively trying to find a better internal build process that would not require this.

ubi9-openjdk-17-runtime.yaml
jmtd: Yes. Let me double-check where that field ends up.

jmtd:
Thank you *very much* for all your review so far!

redhat/ubi9-openjdk-17-runtime.yaml
jerboaa: Got it.

modules/s2i/core/bash/v9/artifacts/opt/jboss/container/s2i/core/s2i-core
jerboaa: +1

modules/jvm/api/v9/module.yaml
jerboaa: Yes, it looks like it's set in `container_memory` and used by `max_memory` and friends. The latter we should be able to get rid of and use the JVM's `*Percentage` flags instead.

jmtd:
Closing in favour of #296 (although Im keeping an eye on the unresolved conversations in this one so I can be sure they are resolved there)

PR #289, opened 2022-05-11
Adjust OpenJDK 17 OSBS package exception

jmtd:
In line with the other images.

Without this, at present, Im seeing build failures due to the presence of some unsigned RPMs in the buildroot, which are totally unrelated to the containers (jigdo for example). This is a transient issue (those RPMs will likely go away or get signed) but this is an unnecessary sensitivity to the issue and a deviation from the other containers.

(JIRA forthcoming)

### DIFF
--- a/redhat/ubi8-openjdk-17.yaml
+++ b/redhat/ubi8-openjdk-17.yaml
@@ -3,7 +3,10 @@ osbs:
     container:
       compose:
         pulp_repos: true
-        packages: []
+        packages:
+        - java-17-openjdk
+        - java-17-openjdk-devel
+        - java-17-openjdk-headless
         signing_intent: release
   repository:
     name: containers/openjdk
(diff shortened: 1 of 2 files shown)


redhat/ubi8-openjdk-17.yaml
```diff
@@ -3,7 +3,8 @@ osbs:
     container:
       compose:
         pulp_repos: true
-        packages: []
+        packages:
+        - java-17-openjdk-headless
```
jerboaa, line 9: AFAIK, for the compose to work it needs to list all installed packages. For `ubi8/openjdk-17` this should be `java-17-openjdk{,-devel,-headless}`.

jmtd: Ah yes I believe you are right, good catch!

jerboaa (approved):

PR #291, opened 2022-05-11
[OPENJDK-779] Drop all openj9 imagestreams

yselkowitz:
With the deprecation of OpenJ9 images, all such references should be removed from the java imagestreams.

https://issues.redhat.com/browse/OPENJDK-779

- [x] A JIRA issue must exist in the OPENJDK project at issues.redhat.com
- [x] Pull Request title should be prefixed with the JIRA issue: `[OPENJDK-XYZ] Subject`
- [x] Pull Request contains hyperlink to the JIRA issue

### DIFF
--- a/templates/image-streams-ppc64le.json
+++ b/templates/image-streams-ppc64le.json
@@ -300,46 +300,6 @@
                         }
                     },
                     {
-                        "name": "openj9-8-el8",
-                        "annotations": {
-                            "openshift.io/display-name": "OpenJ9 1.8.0 (RHEL 8)",
-                            "description": "Build and run Java applications using Maven and OpenJ9 1.8.0.",
-                            "iconClass": "icon-rh-openj9",
-                            "tags": "builder,java,openj9",
-                            "supports": "java:8,java",
-                            "sampleRepo": "https://github.com/jboss-openshift/openshift-quickstarts",
-                            "sampleContextDir": "undertow-servlet",
-                            "version": "8"
-                        },
-                        "referencePolicy": {
-                            "type": "Local"
-                        },
-                        "from": {
-                            "kind": "DockerImage",
-                            "name": "registry.redhat.io/openj9/openj9-8-rhel8:latest"
-                        }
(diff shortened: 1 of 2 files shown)


jmtd:
Hi @yselkowitz , nice to see you, thanks for the PR! I'd attempted this too, in #271, but hadn't merged because I had an open question: What happen if someone has an OpenShift instance with running pods based on these images, and then they import our updated ImageStream metadata that deletes those images?

I remembered that being raised as an issue when removing older image versions from imagestreams but I couldn't remember the specifics.

Without having answered that question, you raising this PR has at least reduced my concern around that. Perhaps there's no problem here?

yselkowitz:
We have been consistently removing unsupported ImageStreamTags (iow versions) from the cluster-samples-operator with each version of OCP since at least 4.6 (possibly earlier, but I wasn't involved then).

I believe the answer to your question wrt removals is "nothing"; images already built will not change, but no new builds will be able to use the removed IST.  However, the one tricky part of this what to do with `java:8` on s390x, as OpenJDK 8 is practically unusable there.  Here I switched it from openj9 to openjdk -- which *might* affect existing builds, depending on how they are configured -- but I'm wondering if it shouldn't just be removed.  (`java:11` is partially in the same boat, except that OpenJDK 11 is actually functional on s390x.)  Note that both `java:8` and `java:11` have been deprecated (and therefore hidden on the web console) since 4.6 when we added ubi8 alongside ubi7.  So this is one point that we should discuss and get in sync between our PRs, but the straight-up removals are all safe.

yselkowitz:
Went ahead and dropped `java:8` and `java:11` for s390x.

jmtd:
Thank you @yselkowitz can you re-target this at the `release` branch please?
(I think I might propose a PR to remove the `templates/` directory in the `develop` branch since nothing consumed it, buit also, we might move the metadata to a separate repository)

yselkowitz:
Resubmitted to release in #295.

PR #294, opened 2022-05-16
Revert Prometheus module change, singleton-jdk for JDK8

jmtd:
There are some issues with the new Prometheus packages in RHEL 8.6, so we
need to back this out to continue to build images.

In RHEL 8.6, prometheus-jmx-exporter depends upon
prometheus-jmx-exporter-jdk-binding, which is provided only by
prometheus-jmx-exporter-openjdk11, which pulls in OpenJDK 11.

Use singleton-jdk to clean up JDK11.

<https://issues.redhat.com/browse/OPENJDK-777>


jerboaa (approved):
Unfortunate, but yes. Go ahead. Hopefully this will be fixed soon.

PR #297, opened 2022-05-23
UBI9 images (alternative approach)

jmtd:
Implement UBI9 images. This is structured as changing the target branch to be exclusively for the UBI9 images. The target is set to `develop` at the moment so that GitHub produces useful diffs in the PR, but ultimately I think we should push this to a new branch.

The PR is structured as follows

1. some pre-ubi9 clean up work, which should be merged into `develop` in #285 and #300 before this branch is merged.
2. re-purposing the UBI8 image descriptors for UBI9 and deleting the rest.
3. removing as much deprecated content as possible
4. simplifying the module structure by merging and renaming modules

The 4th point is perhaps the hardest to review/express what's going on. Here's a graph of the module structure in `develop` right now:

ubi9-before

I've tried to find leaf nodes that have only one parent, and where possible, merged that into the parent. In most cases that means merging "foo.bash" and "foo.api", a separation that was originally designed to permit modules being written in other languages like Python, although we won't want that anymore since we don't want Python in the images at all. 

The module structure in this PR now looks like this:

ubi9-fork-after

(I think there's room for further improvement).

My TODO for this branch:

 - [ ] consider merging java.s2i.bash, maven.s2i, s2i.core,  and possibly maven.default together
 - [ ] consider a flag to disable passing all script-generated JVM parameters
 - [ ] can we get rid of rsync?
 - [ ] revisit the image `description:`
 - [ ] triple check other default JVM parameters: -XX:MaxRAMPercentage=80 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10 -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:+ExitOnOutOfMemoryError
 - [ ] figure out whether to remove one of S2I_ENABLE_INCREMENTAL_BUILDS and MAVEN_CLEAR_REPO


jmtd:
Hi @jerboaa , although I've still got a TODO list for this (in the first comment), IMHO they aren't blockers and think it's probably ready for review (reviewer is welcome to overrule me!). I will aim to address them in a separate PR.

README.adoc
```diff
@@ -6,55 +6,44 @@ These are the OpenJDK Source to Image (S2I) images for Red Hat OpenShift.
 
 ### From https://access.redhat.com/containers/[Red Hat Container Catalog]
 
-The UBI8-based OpenJDK images are available under the terms of the
-link:https://www.redhat.com/en/about/red-hat-end-user-license-agreements#UBI[UBI EULA]:
+The UBI9-based OpenJDK images are available under the terms of the
+link:https://www.redhat.com/en/about/red-hat-end-user-license-agreements#UBI[UBI EULA.
```
jerboaa: Missing closing `]` intentional?

jerboaa (approved):
LGTM as far as I can tell.

README.adoc
jmtd: Good catch thanks

PR #298, opened 2022-05-25
[OPENJDK-777] Adjust Prometheus modules for 8.6 RPM scheme

jmtd:
The required prometheus packages should now be available. This
is a second attempt at #288.

<https://issues.redhat.com/browse/OPENJDK-777>


jerboaa (commented):
Remove ` - name: jboss.container.java.singleton-jdk` from the `ubi8-openjdk-8.yaml` descriptor? GHA seems to fail because of it.

jmtd:
> Remove - name: jboss.container.java.singleton-jdk from the ubi8-openjdk-8.yaml descriptor? GHA seems to fail because of it.

Thanks yes - that's gone now.

PR #300, opened 2022-05-27
Pre ubi9 improvements

jmtd:
These are some miscellaneous image source improvements that I made during work on ubi9 which I think we should apply to the current `develop` branch, as they might ease future maintenance of the other images too.


jerboaa (approved):
OK

PR #301, opened 2022-06-01
Remove vestiges of $HOME/passwd

jmtd:
This was used as a hack when we needed to use NSS wrapper for backwards
compatibility with OCP 3.11.


jerboaa (approved):
+1

PR #302, opened 2022-06-01
[OPENJDK-78] Re-work JAVA_OPTS to override script-calculated java options

jmtd:
https://issues.redhat.com/browse/OPENJDK-78

This is a first attempt at a way of short-circuiting all Java options
generation.

### DIFF
--- a/modules/jvm/module.yaml
+++ b/modules/jvm/module.yaml
@@ -16,10 +16,13 @@ envs:
 - name:         JBOSS_CONTAINER_JAVA_JVM_MODULE
   value:        /opt/jboss/container/java/jvm
 - name:         JAVA_OPTS
-  description:  JVM options passed to the `java` command.
+  description:  If set, this prevents all default and generated options (such
+    as JVM and GC tuning, proxy settings, etc.) from being applied. Instead the
+    user should specify exactly the `java` options they want in this variable.
   example:      "-verbose:class"
 - name:         JAVA_OPTS_APPEND
-  description:  User specified Java options to be appended to generated options in JAVA_OPTS.
+  description:  User specified Java options to be appended to the generated options.
+    This variable has no effect if `JAVA_OPTS` has been defined.
   example:      "-Dsome.property=foo"
 - name: JAVA_MAX_MEM_RATIO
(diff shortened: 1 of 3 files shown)


jmtd:
I was thinking we needed to either continue to support `JAVA_OPTS` or `JAVA_OPTS_APPEND` in addition to the PR in its current state. Then, I wondered, what purpose does `JAVA_OPTS` serve at the moment anyway? Perhaps we should just let *that* override all auto-calculated options, and save introducing another variable.

jmtd:
The PR now better reflects how I remember `JAVA_OPTS` behaving in the dim and distant past.

jerboaa (commented):
We should have a test which verifies that if `JAVA_OPS` and `JAVA_OPTS_APPEND` are both specified, then only `JAVA_OPTS` are being added. I.e. `JAVA_OPTS_APPEND` is an override for the default options. `JAVA_OPTS` are a replacement.

jmtd:
Good idea: I've just pushed one.

modules/jvm/module.yaml
```diff
@@ -16,10 +16,12 @@ envs:
...
   value:        /opt/jboss/container/java/jvm
 - name:         JAVA_OPTS
-  description:  JVM options passed to the `java` command.
+  description:  If set, this prevents all default and generated options (such
+    as JVM and GC tuning, proxy settings, etc.) from being applied. Instead the
+    user should specify exactly the `java` options they want in this variable.
   example:      "-verbose:class"
 - name:         JAVA_OPTS_APPEND
-  description:  User specified Java options to be appended to generated options in JAVA_OPTS.
+  description:  User specified Java options to be appended to the generated options.
```
jerboaa, line 24: Please add that this variable has no effect if `JAVA_OPTS` is set in the description.

jerboaa (approved):
LGTM. Please clarify the description of `JAVA_OPTS_APPEND`.

modules/jvm/module.yaml
jmtd: I don't mind doing so, and I will. I originally decided not to because I wasn't also going to annotate all the other variables that are disabled by specifying `JAVA_OPTS`, but, due to their similar name and behaviour I can see that this one should be treated differently.

jmtd:
Again, the failing test here passes locally and I think is sporadic on GH actions.

PR #306, opened 2022-06-27
[OPENJDK-91] Rename jboss user to 'default'

jmtd:
<https://issues.redhat.com/browse/OPENJDK-91>

The use of "jboss" for the runtime user is sometimes confusing for users
who are not using a JBoss product. Rename it to "default", in line with
some other UBI images. This also changes the $HOME path. We retain the
uid 185.

I'm not sure how disruptive this might be!

### DIFF
--- a/modules/jdk/11/configure.sh
+++ b/modules/jdk/11/configure.sh
@@ -5,7 +5,7 @@ set -e
 SCRIPT_DIR=$(dirname $0)
 ARTIFACTS_DIR=${SCRIPT_DIR}/artifacts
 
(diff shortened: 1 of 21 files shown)


jerboaa (approved):
OK. I take it this will only be used by the UBI 9 images?

jmtd:
> OK. I take it this will only be used by the UBI 9 images?

Yeah, I decided against doing this when it was first proposed due to the backwards compatibility issues. I think a RHEL major version bump / new image name is the only opportunity to do it. This came up again last week so I thought, now or never. Still a little nervous about it though.

PR #312, opened 2022-07-13
Lower the default MaxRAMPercentage for Maven JVMs

jmtd:
https://issues.redhat.com/browse/OPENJDK-559

PR #320, opened 2022-08-23
[Openjdk 559] max ram percentage

jmtd:
This is a repeat of #285, with the maven fix from #312 included.

https://issues.redhat.com/browse/OPENJDK-559


jerboaa (approved):
LGTM. Please check that GHA passes once the underlying issue is fixed.

PR #325, opened 2022-08-31
[OPENJDK-986] Remove dejavu-sans-mono-fonts from test

jmtd:
This specific test is intended to catch packages that are not covered by
the UBI EULA leaking into built images. The package
dejavu-sans-mono-fonts has joined the UBI set since this test was
written.

https://issues.redhat.com/browse/OPENJDK-986


jerboaa (approved):
LGTM

PR #328, opened 2022-10-20
[OPENJDK-1065] Add PNC Metadata to artifacts

jmtd:
This causes OSBS-builds of these containers to consume the artifacts from an internal PNC instance.

The 'md5' key is not valid when the pnc keys are present. Remove it, but add two tests that check the md5 checksum of the JARs inside an image.

https://issues.redhat.com/browse/OPENJDK-1065

### DIFF
--- a/.github/workflows/ubi8-openjdk-11-runtime.yml
+++ b/.github/workflows/ubi8-openjdk-11-runtime.yml
@@ -2,7 +2,7 @@ name: OpenJDK 11 Runtime Image CI
 on: [push, pull_request]
 env:
   LANG: en_US.UTF-8
-  CEKIT_VERSION: 3.11.0
+  CEKIT_VERSION: 4.3.0
   S2I_URI: https://github.com/openshift/source-to-image/releases/download/v1.3.1/source-to-image-v1.3.1-a5a77147-linux-amd64.tar.gz
 jobs:
   openjdkci:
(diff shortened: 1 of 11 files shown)


jmtd:
Review tip:
check internal Brew for taskinfo?taskID=48451353. That was built from an in-development tree where the jolokia portion of this was done but the prometheus was not. Notice "user_warnings = No source-url found for…"

taskinfo?taskID=48452279 was built after the prometheus changes in this PR. Notice no warnings in the Result section.

jmtd:
This breaks GitHub Actions, because the currently configured cekit version for those is 3.11.0, and PNC support was added in 3.12.0. However, we have no GitHub CI coverage for the RHEL7-based images anyway. I coudl push a commit that bumps the cekit version so the lights go green, but we should consider the more drastic removal of GitHub CI for the rhel7 branch.

jerboaa:
> This breaks GitHub Actions, because the currently configured cekit version for those is 3.11.0, and PNC support was added in 3.12.0. However, we have no GitHub CI coverage for the RHEL7-based images anyway. I coudl push a commit that bumps the cekit version so the lights go green, but we should consider the more drastic removal of GitHub CI for the rhel7 branch.

Please push to newer cekit in GHA. It's a CI only change and should do no harm, right?. I think the latest version these days is cekit `4.x` anyway.

jmtd:
> Please push to newer cekit in GHA. It's a CI only change and should do no harm, right?. I think the latest version these days is cekit 4.x anyway.

Done (let's see what GHA makes of it). If it works commit is suitable for cherry-picking over to the other branches, and I'll do that.

jerboaa (approved):

PR #329, opened 2022-11-01
[OPENJDK-1130] revert accidental commits to rhel7 images

jmtd:
https://issues.redhat.com/projects/OPENJDK/issues/OPENJDK-1130

These commits were made on what was the "develop" branch intended for the ubi8 images, but we forked the rhel7 branch *afterwards*.


jerboaa (approved):
OK. These checksum failures for prometheus-jmx-agent need investigating:

```
  features/image/prometheus.feature:33  Ensure the Prometheus Agent JAR has the expected checksum
  File "/home/runner/cekit4.3.0/lib/python3.6/site-packages/cekit/builder.py", line 27, in execute
```

jmtd:
The md5 hash in the unit test is correct for the artefact we ship in RHEL7 :

```
-- md5: 8b3af39995b113baf35e53468bad7aae
-  target: jmx_prometheus_javaagent-0.3.2.redhat-00003.jar
```

I've just confirmed a scratch build with the new pnc stuff pulled in the right artefacts.

But the hash is not correct for UBI8 since we updated prometheus-jmx-exporter-0.12.0-8.el8_6.noarch
(it is now 1a9b7eb478c72028ca91e1bf6e40b48f).

So,

1) this is further evidence we should just turn off GHA for the rhel7 branch IMHO
2) we need to update the hash - but in the ubi8 branch, not here

PR #331, opened 2022-12-08
[OPENJDK-1334] Move pkg-update to end of container build

jmtd:
This causes the "microdnf update" to take place after all the RPMs that are explicitly or implicitly installed by OpenJDK, Maven, etc have been installed.

This works around the issue described in RHELDST-15034.

https://issues.redhat.com/browse/OPENJDK-1334


jerboaa (approved):
OK.

PR #332, opened 2022-12-22
[OPENJDK-1335] microdnf update weak deps

jmtd:
Part 1 of a 2-commit PR: I want to see CI red lights for this commit before pushing the fix.

This commit adds a test for some RPMs that are installed as transitive weak dependencies via microdnf when it is not configured with install_weak_deps=0.

https://issues.redhat.com/browse/OPENJDK-1335?filter=12398815

### DIFF
--- a/modules/util/pkg-update/execute.sh
+++ b/modules/util/pkg-update/execute.sh
@@ -4,7 +4,7 @@ set -euo pipefail
 for candidate in yum dnf microdnf; do
     if command -v "$candidate"; then
         mgr="$(command -v "$candidate")"
-        "$mgr" update -y
+        "$mgr" update --setopt=install_weak_deps=0 --setopt=tsflags=nodocs -y
         "$mgr" -y clean all
         exit
     fi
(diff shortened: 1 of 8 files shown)


jmtd:
The new test passed for ubi8/openjdk-8 but shouldn't have. I think it's a race condition, due to using "available container log". I'm going to try just "container log". This unfortunately means a 30s wait per step in the scenario, but "correct" trumps "fast".

modules/util/pkg-update/execute.sh
```diff
@@ -4,7 +4,7 @@ set -euo pipefail
 for candidate in yum dnf microdnf; do
     if command -v "$candidate"; then
         mgr="$(command -v "$candidate")"
-        "$mgr" update -y
+        "$mgr" update --setopt=install_weak_deps=0 -y
```
jerboaa: Please also set `--setopt=tsflags=nodocs` that's what regular install commands have as well.

jerboaa (approved):
Looks good. Thanks!

PR #337, opened 2023-01-31
[Openjdk-559] max ram percentage default

jmtd:
https://issues.redhat.com/browse/OPENJDK-559

Defer the change of max heap size to 80%, revert to previous default of 50%.

### DIFF
--- a/modules/jvm/api/module.yaml
+++ b/modules/jvm/api/module.yaml
@@ -14,7 +14,7 @@ envs:
   description: User specified Java options to be appended to generated options in JAVA_OPTS.
   example: "-Dsome.property=foo"
 - name: JAVA_MAX_MEM_RATIO
-  description:  Specify the maximum heap memory. Corresponds to the JVM argument `-XX:MaxRAMPercentage`. The default is `80.0` which means 80% of the available memory. You can disable this mechanism by setting the value to `0`.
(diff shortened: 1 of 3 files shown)


modules/jvm/api/module.yaml
```diff
@@ -14,7 +14,7 @@ envs:
   description: User specified Java options to be appended to generated options in JAVA_OPTS.
   example: "-Dsome.property=foo"
 - name: JAVA_MAX_MEM_RATIO
-  description:  Specify the maximum heap memory. Corresponds to the JVM argument `-XX:MaxRAMPercentage`. The default is `80.0` which means 80% of the available memory. You can disable this mechanism by setting the value to `0`.
+  description:  Specify the maximum heap memory. Corresponds to the JVM argument `-XX:MaxRAMPercentage`. The default is `50.0` which means 80% of the available memory. You can disable this mechanism by setting the value to `0`.
```
jerboaa: `s/which means 80%/which means 50%/`

jerboaa (commented):
Just a type. Good otherwise.

modules/jvm/api/module.yaml
jmtd: Thanks, embarrassing :( fixed

PR #338, opened 2023-02-09
[Openjdk-1464] upgrade to  maven 3.8

jmtd:
https://issues.redhat.com/browse/OPENJDK-1464

I'm not sure when to release this, whether we need to give people some warning. Gathering further information.

The approach I've taken assumes we don't need to mix maven versions between images in the ubi9 branch.

Note simplification of module versioning scheme -- see comment in the second commit in series for a fuller explanation.

All current behave tests pass.


jmtd:
Test builds:

* quay.io/jdowland/ubi9-openjdk-17:OPENJDK-1464
* quay.io/jdowland/ubi9-openjdk-11:OPENJDK-1464

jerboaa (commented):
Why not call it by the maven version number `3.8`? It seems counter-intuitive to see `jboss.container.maven` => `version: 11` and `version: 17` both pulling in module version 3.8.

jerboaa:
> Why not call it by the maven version number `3.8`? It seems counter-intuitive to see `jboss.container.maven` => `version: 11` and `version: 17` both pulling in module version 3.8.

Nevermind. It's because of the `maven-openjdkXX` package, which is JDK specific. I'd propose to call it `3.8.11` and `3.8.17`, though.

jerboaa (approved):

PR #339, opened 2023-02-15
Bump version for next release

jmtd:
1.13 has shipped


jmtd:
Check incremental builds cache .m2 is an unreliable test

this is the only test failure, I think it's unrelated. Will track that in another issue.

PR #344, opened 2023-03-15
[Openjdk-1551] JAVA_APP_NAME quoting/interpolation issues

jmtd:
https://issues.redhat.com/browse/OPENJDK-1551

The reporter specifically highlighted two issues

 *   JAVA_APP_NAME=foo means the process name will be 'foo' (with the single quotes included
 *   JAVA_APP_NAME=foo bar will cause the container to crash loop with exec: bar': not found

Both should now be addressed, and there's a test for the latter.

Remove a stray string from an earlier test (pebkac)

### DIFF
--- a/modules/run/artifacts/opt/jboss/container/java/run/run-java.sh
+++ b/modules/run/artifacts/opt/jboss/container/java/run/run-java.sh
@@ -182,13 +182,6 @@ get_classpath() {
   echo "${cp_path}"
 }
 
-# Set process name if possible
-get_exec_args() {
-  if [ "x${JAVA_APP_NAME}" != x ]; then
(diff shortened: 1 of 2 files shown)


jerboaa (approved):
Looks good.

jmtd:
Test images:
* quay.io/jdowland/ubi9-openjdk-11:OPENJDK-1551
* quay.io/jdowland/ubi9-openjdk-11-runtime:OPENJDK-1551
* quay.io/jdowland/ubi9-openjdk-17:OPENJDK-1551
* quay.io/jdowland/ubi9-openjdk-17-runtime:OPENJDK-1551

PR #348, opened 2023-04-05
[OPENJDK-1549] clean environment when invoking Maven

### LINKED ISSUE #343: MAVEN_ARGS is known by mvn starting 3.9
As we can see there: https://issues.apache.org/jira/browse/MNG-7193 MAVEN_ARGS is natively supported by maven 3.9
Thia has an impact on s2i. This line for example: https://github.com/jboss-container-images/openjdk/blob/ubi8/modules/maven/default/artifacts/opt/jboss/container/maven/default/maven.sh#L85 would have to be modified to not set $MAVEN_ARGS in the command line (it would duplicate the args).

jmtd:
<https://issues.redhat.com/browse/OPENJDK-1549>

This is a proposed solution to the forthcoming issue that Maven (from
version 3.9) will start to honour the environment variable MAVEN_ARGS
(see #343).

The proposal is we clean the environment before launching Maven, such
that none of the variables we use for configuring the images leak out.

This would break a workflow whereby a user has deliberately defined a
variable for Maven to consume, within S2I.  I think this is unlikely,
but I am seeking feedback from the wider Middleware community.

Similarly (and not implemented in this PR) I think we should consider
the same approach when launching Java. That is, scrub the environment
in the run script prior to invoking /usr/bin/java. If a user defined
environment variables intended to reach their application, this would
break. I suspect this is more likely than in the Maven case, but I am
opening a discussion.

As an implementation note: I needed to define a Maven POM in order to
write a test for this, and I needed somewhere to store that. I didn't
think <https://github.com/jboss-openshift/openshift-quickstarts>  was
an appropriate place, since the test POM is not a quickstart, and now
we have archived that repository. So, I added it to this repository.
Due to the chicken-and-egg nature of needing to reference the POM by
URI prior to this being merged, the test references my fork instead.
This could be corrected in a follow-up PR.

Fixes: #343.

### DIFF
--- a/modules/maven/default/artifacts/opt/jboss/container/maven/default/maven.sh
+++ b/modules/maven/default/artifacts/opt/jboss/container/maven/default/maven.sh
@@ -69,16 +69,16 @@ function maven_build() {
   local goals=${2:-package}
   log_info "Performing Maven build in $build_dir"
 
-  pushd $build_dir &> /dev/null
+  # subshell so we can exec mvn with a clean environment (OPENJDK-1549)
+  (
+    cd $build_dir
 
-  log_info "Using MAVEN_OPTS ${MAVEN_OPTS}"
-  log_info "Using $(mvn $MAVEN_ARGS --version)"
-  log_info "Running 'mvn $MAVEN_ARGS $goals'"
-  # Execute the actual build
-  mvn $MAVEN_ARGS $goals
-  
-  popd &> /dev/null
-  
+    log_info "Using MAVEN_OPTS ${MAVEN_OPTS}"
+    log_info "Using $(mvn $MAVEN_ARGS --version)"
+    log_info "Running 'mvn $MAVEN_ARGS $goals'"
+
+    exec -c mvn $MAVEN_ARGS $goals
+  )
 }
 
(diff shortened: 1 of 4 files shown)


jerboaa (approved):

jerboaa:
> Similarly (and not implemented in this PR) I think we should consider
> the same approach when launching Java.

I'm not sure this would be a good idea. The runtime system of a container pretty much recommends using environment variables to get some state set. Things like `LANG` etc. in our Dockerfile come to mind. They'd all be gone? Sounds like a very breaking change.

jmtd:
> I'm not sure this would be a good idea. The runtime system of a container pretty much recommends using environment variables to get some state set. Things like LANG etc. in our Dockerfile come to mind. They'd all be gone? Sounds like a very breaking change.

Yes, perhaps carte blanche is a bad idea. I wonder about a white/black list approach. Perhaps we could explicitly unset the variables that we define for our own purposes (`JAVA_OPTS_APPEND` etc.). Or just not worry about it for the application: we haven't hit a problem on that side after all.

jmtd:
I think I'm going to self-NACK this approach: it's too risky that we break a customer wishing to pass an Env through to Maven on purpose.

PR #349, opened 2023-05-04
[OPENJDK-1673] use cp -p for s2i deployments

jmtd:
The test is preliminary, I guess it would be good to switch it to the quarkusio organisation rather than @jerboaa 's fork. I'm not sure if there's an appopriate ref in the mainline (in place of quickstart-2.16-s2i-cds); is any of the CDS stuff landed? Thanks

https://issues.redhat.com/browse/OPENJDK-1673


jerboaa:
> The test is preliminary, I guess it would be good to switch it to the quarkusio organisation rather than @jerboaa 's fork. I'm not sure if there's an appopriate ref in the mainline (in place of quickstart-2.16-s2i-cds); is any of the CDS stuff landed? 

+1, but we aren't there yet. Once we have something we know works I'll propose the patch to the quarkus quickstarts repo. Then we can switch.

jerboaa (approved):
Seems OK.

PR #352, opened 2023-05-05
Move GHA runners to ubuntu-22.04

jmtd:
This is a blind change, I don't know whether further work will be needed to update the actions for Ubuntu 22.04. Hopefully the runs for this PR will shed some light.

PR #355, opened 2023-05-19
Update GHA runners to ubuntu-20.04

jmtd:
UBI8 version of #352 .


jerboaa:
The new quarkus CDS test from #351 fails on JDK 8 with:

```
 Error: RROR:cekit:I0519 11:08:01.110417   29769 sti.go:711] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project getting-started: Fatal error compiling: invalid target release: 11 -> [Help 1]
```

I think we should only run that test when on JDK 11+. Also, we should investigate if that test does what it's supposed to, since I find it curious it currently passes.

jerboaa (approved):
On the basis that test now run again, approved. Please follow up with the test issue. Thanks!

PR #359, opened 2023-06-14
WIP [OPENJDK-1956] assemble-runtime S2I script

jmtd:
https://issues.redhat.com/browse/OPENJDK-1956

This is not ready so I am not marking it as for review, yet, but this is a MVP that lets the s2i runtime workflow function.


jmtd:
we need to add `/usr/local/s2i/run` or otherwise address its absence. One option is to split the modules/s2i/bash up.

jmtd:
Based on discussions, our current approach is not to use an S2I/runtime workflow to deliver this anymore, so I'm closing the two WIP PRs for now.

PR #366, opened 2023-07-21
[OPENJDK-2033] OpenJDK runtime images cannot overwrite JAVA_APP_DIR

ciis0:
[OPENJDK-2033](https://issues.redhat.com/browse/OPENJDK-2033)

without this `run-java.sh` always overwrites the user-provided `JAVA_APP_DIR` with the one from `run-env.sh`.

- [x] A JIRA issue must exist in the OPENJDK project at issues.redhat.com
- [x] Pull Request title should be prefixed with the JIRA issue: `[OPENJDK-XYZ] Subject`
- [x] Pull Request contains hyperlink to the JIRA issue

### DIFF
--- a/modules/run/artifacts/opt/jboss/container/java/run/run-env.sh
+++ b/modules/run/artifacts/opt/jboss/container/java/run/run-env.sh
@@ -1,4 +1,6 @@
 #!/bin/sh
 
-# Set the application dir to the S2I deployment dir
-JAVA_APP_DIR=/deployments
+# Default the application dir to the S2I deployment dir
+if [ -z "$JAVA_APP_DIR" ]
+then JAVA_APP_DIR=/deployments
+fi
--- a/modules/run/artifacts/opt/jboss/container/java/run/run-java.sh
+++ b/modules/run/artifacts/opt/jboss/container/java/run/run-java.sh
@@ -245,6 +245,7 @@ startup() {
   procname="${JAVA_APP_NAME-java}"
 
   log_info "exec -a \"${procname}\" java $(get_java_options) -cp \"$(get_classpath)\" ${args} $*"
+  log_info "running in $PWD"
   exec -a "${procname}" java $(get_java_options) -cp "$(get_classpath)" ${args} $*
 }
 
--- a/tests/features/java/runtime.feature
+++ b/tests/features/java/runtime.feature
@@ -67,3 +67,15 @@ Feature: Openshift OpenJDK Runtime tests
     | JAVA_OPTIONS | --show-version |
     Then container log should not contain -XX:MaxRAMPercentage
     And  container log should contain --show-version
+
+  @ubi8
+  Scenario: Check default JAVA_APP_DIR (OPENJDK-2033)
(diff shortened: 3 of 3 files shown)


jmtd:
Hi, thanks for the issue, and the fix! It looks good, I'm going to see if we can get this into our next feature update (to ship in August). We will need the same fix in the ubi9 branch, and I would like to see some behave test coverage for JAVA_APP_DIR behaviour. Would you be prepared to tackle either of those?

ciis0:
I'll look into both. :)

Test like here, yes? https://github.com/jboss-container-images/openjdk/tree/ubi8/tests/features/java, or maybe even in https://github.com/jboss-container-images/openjdk/blob/ubi8/tests/features/java/runtime.feature directly?

jmtd:
Yes thats exactly right, and directly into `runtime.feature` would be a good place. If you enable GitHub actions on your fork, it should run the test suite on any pushes you make. Thanks!

jmtd:
Hi @ciis0 , I think we can probably ship this fix in an update in roughly mid-November. Are you still planning to take a look at the tests? Thanks!

ciis0:
Yes still planning to look into it.

I got lost last time in not having podman instead of docker locally, which breaks S2I tests... (https://github.com/openshift/source-to-image/issues/1082) ^^

but the non-s2i tests work normally with podman, and such a test should be enough for JAVA_APP_DIR tests. :)

ciis0:
I have added tests and a PR for UBI9 now. :)

jmtd (approved):

jmtd:
Thanks for working on this, I am very happy with it. I've approved the changes to merge ASAP.

Right now, we're prepping for the next container release (likely this Month). I'm in a meeting this afternoon where I will see if I can get these on the slate for that release. If so I can merge them now; otherwise I will have to wait until we've got that release out.

sefroberg:
I can see the need for two additional tests here.
1. What happens if the variable is set to just the folder and not the full path?
2. What happens if that folder does not exist in the container?

ciis0:
> I can see the need for two additional tests here.
> 
>     1. What happens if the variable is set to just the folder and not the full path?
> 

I think it will be relative to WORKDIR, but I think I rather would enforce that APP_DIR must start with a /, as a cheap "must be absolute" test, that should be less confusing.

>     2. What happens if that folder does not exist in the container?

I assume the cd into APP_DIR will fail. I will add a check for existence with an appropriate error message (and a corresponding test :) ).
I could also add creating the APP_DIR, but I assume that by default will simply fail due to runtime fs permissions.

jmtd:
@ciis0 , I didn't see your comment here before I replied on the ubi9 PR (I tend to start on ubi9). Feel free to cherry-pick and adapt https://github.com/jmtd/redhat-openjdk-container-images/commit/609d67ecff7e031e8c02b14b160eafac97ec4121 over to here!

Personally I don't feel the need for behaviour changes to handle these two situations.

sefroberg:
Looks good to me with the updated behave tests.

PR #369, opened 2023-07-21
[OPENJDK-2009] let JAVA_OPTS override all tuning parameters (ubi8)

jmtd:
https://issues.redhat.com/browse/OPENJDK-2009

An adaptation of OPENJDK-78 for UBI9: make JAVA_OPTS override all tuning parameters.


jerboaa:
CI doesn't look happy about this :(

jmtd:
This seems to be because JAVA_OPTS gets touched (and defined as empty as opposed to undefined) in the s2i run script, as part of the prometheus/jolokia configuration stuff (which are absent on ubi9). Working on it

modules/s2i/bash/artifacts/usr/local/s2i/run
```diff
@@ -17,14 +17,19 @@ if test -r "${JBOSS_CONTAINER_MAVEN_DEFAULT_MODULE}/scl-enable-maven"; then
     source "${JBOSS_CONTAINER_MAVEN_DEFAULT_MODULE}/scl-enable-maven"
 fi
 
-JAVA_OPTS="${JAVA_OPTS:-${JAVA_OPTIONS}}"
+if [ -z "${JAVA_OPTS+isunset}" ] && [ -n "${JAVA_OPTIONS+isset}" ]; then
+    JAVA_OPTS="$JAVA_OPTIONS"
+fi
```
jerboaa, line 22: Please add a test for `JAVA_OPTIONS=foo` and `JAVA_OPTS` being unset, expecting `foo` being passed to the java command line.

jmtd: I've added that test, and it fails, now to work out why.

jmtd:
Observation/subtlety: the CMD for the builders is `/usr/local/s2i/run` (where this patch is operating) but for the runtime images it's `/opt/jboss/container/java/run/run-java.sh`. Therefore there is a divergent behaviour wrt `JAVA_OPTIONS` (but also, the runtime images have never supported it.)

The runtime images will eventually get `/usr/local/s2i/run`  as part of [OPENJDK-2063](https://issues.redhat.com/browse/OPENJDK-2063).

jerboaa:
> Observation/subtlety: the CMD for the builders is `/usr/local/s2i/run` (where this patch is operating) but for the runtime images it's `/opt/jboss/container/java/run/run-java.sh`. Therefore there is a divergent behaviour wrt `JAVA_OPTIONS` (but also, the runtime images have never supported it.)

That suggests the test should be tagged builder only for now and enabled as part of the runtime images changes. Thoughts?

jmtd:
> That suggests the test should be tagged builder only for now and enabled as part of the runtime images changes. Thoughts?

Agree with the first part: for the second, I might see if we can ensure we never support `JAVA_OPTIONS` for those.

jerboaa (approved):
Seems good. Consider adding `JAVA_OPTIONS => JAVA_OPTS` to your private ubi8 -> ubi9 migration guide notes ;-)

PR #370, opened 2023-07-21
Ensure Quarkus CDS test is not tagged for JDK8

### LINKED ISSUE #356: The new quarkus CDS test fails on JDK 8, shouldn't run on 8
The new quarkus CDS test from #351 fails on JDK 8 with:

```
 Error: RROR:cekit:I0519 11:08:01.110417   29769 sti.go:711] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project getting-started: Fatal error compiling: invalid target release: 11 -> [Help 1]
```

I think we should only run that test when on JDK 11+. Also, we should investigate if that test does what it's supposed to, since I find it curious it currently passes.

_Originally posted by @jerboaa in https://github.com/jboss-container-images/openjdk/issues/355#issuecomment-1554448662_

jmtd:
Fixes #356.

### DIFF
--- a/tests/features/java/openjdk_s2i.feature
+++ b/tests/features/java/openjdk_s2i.feature
@@ -2,6 +2,7 @@
 @ubi8/openjdk-11
 @ubi8/openjdk-17
 Feature: Openshift OpenJDK-only S2I tests
+
   @ubi8/openjdk-8
   Scenario: Check java perf dir owned by jboss (CLOUD-2070)
     Given s2i build https://github.com/jboss-openshift/openshift-quickstarts from undertow-servlet


tests/features/java/openjdk_s2i.feature
```diff
@@ -2,6 +2,7 @@
 @ubi8/openjdk-11
 @ubi8/openjdk-17
 Feature: Openshift OpenJDK-only S2I tests
+
   @ubi8/openjdk-8
   Scenario: Check java perf dir owned by jboss (CLOUD-2070)
```
jerboaa, line 7: Are you sure this is the gist of the patch? If so, please explain why that works?

jmtd: The 'Feature' line has the other image names tagged above it. Other Scenarios later in the file (separated from both the Feature and Scenario lines by whitespace) inherit their tagging from the Feature. 

Without this newline, the 'Feature' line, above, and the 'Scenario' line, below, are considered part of the same stanza, and so the tag `@ubi8/openjdk-8` is applying to the Feature as well, and being inherited by all the other Scenarios (including the CDS one).

jerboaa: Oh boy, what terrible design. Thanks for the explanation.

jerboaa (approved):
OK. Ship this magic!

PR #371, opened 2023-07-26
[OPENJDK-2048] imagestream scripts and updated imagestreams

### LINKED ISSUE #270: ImageStream templates need better management
There is a lot of duplication between records inside each ImageStream template file, and even more across files. Updating to add a single minor version or similar is a lot of work with a lot of room for human error. It would be good to make them as concise as possible, reduce duplication and make it easier to make such adjustments safely somehow. Preprocessing? Switching to YAML?

jmtd:
https://issues.redhat.com/browse/OPENJDK-2048

This PR adds a Python script* which generates the ImageStream template files in `templates/*` using data on version ranges, architecture exceptions, etc. from `data.yaml`.

I've split the PR up into 5 commits to make review easier.

1. First we add a very simple GitHub Action to verify the templates are valid JSON (pass them through jq)
2. The generator script (`templates/generateImageStreamTemplates.py`), its data source (`templates/data.yaml`) and template snippets (`templates/templates/*`) are added. The data source is configured to match the versions that are included in the _current templates_.
3. Commit the result of running the script. There are some differences from the static templates, documented in the commit message.
4. Update `data.yaml` to reflect the current latest shipped version of the containers. (The original templates were out of date, which motivated this PR.)
5. Commit the result of running the script again: New Image versions are appended in the right places

Fixes #270.

_* using the same libraries as Cekit, to maximize the chance other folks might work on this in future._

### DIFF
--- a/.github/workflows/verifyTemplates.yml
+++ b/.github/workflows/verifyTemplates.yml
@@ -1,0 +1,26 @@
+name: Validate ImageStream templates
+
+on:
+  push:
+    branches:
+      - release
+  pull_request:
+    branches:
+      - release
+
+jobs:
+  verify:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        template: 
+          - community-image-streams.json
+          - image-streams.json
+          - image-streams-aarch64.json
+          - image-streams-ppc64le.json
+          - image-streams-s390x.json
+          - runtime-image-streams.json
+    steps:
+    - uses: actions/checkout@v2
+    - name: run jq
+      run: jq < templates/${{ matrix.template }}
--- a/templates/community-image-streams.json
+++ b/templates/community-image-streams.json
@@ -96,6 +96,82 @@
                             "kind": "DockerImage",
                             "name": "registry.access.redhat.com/ubi8/openjdk-8:1.12"
                         }
+                    },
+                    {
+                        "name": "1.13",
+                        "annotations": {
+                            "openshift.io/display-name": "Red Hat OpenJDK 8 (UBI8)",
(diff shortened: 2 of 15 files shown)


.github/workflows/verifyTemplates.yml
```diff
@@ -0,0 +1,24 @@
+name: Validate ImageStream templates
+
+on:
+  push:
+    branches: [ "release", "OPENJDK-2048-imagestream-scripts" ]
+  pull_request:
+    branches: [ "release", "OPENJDK-2048-imagestream-scripts" ]
```
jerboaa: `OPENJDK-2048-imagestream-scripts` branch mention shouldn't be part of the change? Verification could be done on a derivative branch, right?

jerboaa (approved):
OK, with a minor comment.

.github/workflows/verifyTemplates.yml
jmtd: Once merged we definitely don't need mention of this branch. And we might not now, either. I wasn't sure (still not) whether `on.pull_request.branches` refers to the source (`OPENJDK-2048-imagestream-scripts`) or target branch (`release`). If the former, we need this pre-merge to run the job in the PR itself. But if I look at the docs more closely (I spent a lot of time with them last week for some recent `jlink-dev` changes) I can clarify that.

jmtd: I've reviewed the docs, and the branch specified here is the target branch, so `release` is valid and `OPENJDK-2048-imagestream-scripts` superfluous. I've updated the PR accordingly.

PR #377, opened 2023-08-08
deliberately break a JSON template to test GHA

jmtd:
This is simply a test that the GH action to validate the templates is validating the copy of the templates in the PR branch, rather than the target branch. This PR deliberately breaks one of them. This should not be merged.


jmtd:
Success: the actions ran on the PR file content

PR #384, opened 2023-08-15
[OPENJDK-2094] explicit tzdata-java dependency in container

jmtd:
This is a temporary measure to mitigate a regression in the OpenJDK 11 RPMs for the 2023-07 CPU update (which accidentally dropped the dependency on tzdata-java).

https://issues.redhat.com/browse/OPENJDK-2094


sefroberg:
Looks good to me.

PR #387, opened 2023-08-25
Openjdk 2034 another test S2I_TARGET_DEPLOYMENTS_DIR and JAVA_APP_DIR

jmtd:
https://issues.redhat.com/browse/OPENJDK-2034

    The S2I process puts the build application in /deployments unless
    overridden by S2I_TARGET_DEPLOYMENTS_DIR. Add a test to ensure that
    S2I_TARGET_DEPLOYMENTS_DIR and JAVA_APP_DIR set to the same value
    work as intended.

### DIFF
--- a/tests/features/java/runtime.feature
+++ b/tests/features/java/runtime.feature
@@ -43,26 +43,26 @@ Feature: Openshift OpenJDK Runtime tests
   Then container log should not contain exec: bar': not found
 
   @ubi9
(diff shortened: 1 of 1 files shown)


sefroberg (approved):
I do not think you need to update the ticket number on the previously created tests but I also do not think it should break anything. The new test looks correct. Please make sure it runs as you expect in the test run. 

Looks good to me.

PR #392, opened 2023-09-06
[OPENJDK-2145] Update/add usage: labels to UBI8 images

jmtd:
https://issues.redhat.com/browse/OPENJDK-2145

Add a usage: label to JDK11-runtime and JDK8-runtime, pointing at <https://jboss-container-images.github.io/openjdk/>, which is probably the best option for the time being. We don't have specific Red Hat customer documentation for runtime images.

Alter the usage: label for JDK17 to
<https://jboss-container-images.github.io/openjdk/>, as the old value was for OpenShift 3.x era documentation.

Adjust the usage: label for JDK11 to a fixed URI for the same document (It changed at some point in the past and we didn't notice sooner.)

For all UBI8 images, additionally define a second label org.opencontainers.image.documentation with the same value, following the OpenContainers Spec, Pre-Defined Annotation Keys: <https://specs.opencontainers.org/image-spec/annotations/>

### DIFF
--- a/ubi8-openjdk-11.yaml
+++ b/ubi8-openjdk-11.yaml
@@ -19,7 +19,7 @@ labels:
 - name: "com.redhat.component"
   value: "openjdk-11-ubi8-container"
 - name: "usage"
-  value: &docs "https://access.redhat.com/documentation/en-us/openjdk/11/html/using_source-to-image_for_openshift_with_red_hat_build_of_openjdk_11/index"
+  value: &docs "https://jboss-container-images.github.io/openjdk/"
 - name: "com.redhat.license_terms"
   value: "https://www.redhat.com/en/about/red-hat-end-user-license-agreements#UBI"
 - name: "org.opencontainers.image.documentation"
--- a/ubi8-openjdk-11.yaml
+++ b/ubi8-openjdk-11.yaml
@@ -19,9 +19,11 @@ labels:
 - name: "com.redhat.component"
   value: "openjdk-11-ubi8-container"
 - name: "usage"
-  value: "https://access.redhat.com/documentation/en-us/openjdk/11/html/using_openjdk_11_source-to-image_for_openshift/index"
+  value: &docs "https://access.redhat.com/documentation/en-us/openjdk/11/html/using_source-to-image_for_openshift_with_red_hat_build_of_openjdk_11/index"
 - name: "com.redhat.license_terms"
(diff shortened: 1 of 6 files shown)


ubi8-openjdk-11.yaml
```diff
@@ -19,9 +19,11 @@ labels:
 - name: "com.redhat.component"
   value: "openjdk-11-ubi8-container"
 - name: "usage"
-  value: "https://access.redhat.com/documentation/en-us/openjdk/11/html/using_openjdk_11_source-to-image_for_openshift/index"
+  value: &docs "https://access.redhat.com/documentation/en-us/openjdk/11/html/using_source-to-image_for_openshift_with_red_hat_build_of_openjdk_11/index"
```
jerboaa: I'd rather keep this consistent with other images. I.e. point to https://jboss-container-images.github.io/openjdk/ instead. It's up-to-date and in-sync with the releases. The product doc could gain a reference link to the upstream doc as well.

jmtd: I'd be happy doing that. And also linking to the product docs from the upstream docs (we link from the README already).

sefroberg: I agree with @jerboaa . I would like to see consistency across the documentation.

ubi8-openjdk-8.yaml
```diff
@@ -19,9 +19,11 @@ labels:
 - name: "com.redhat.component"
   value: "openjdk-8-ubi8-container"
 - name: "usage"
-  value: "https://access.redhat.com/documentation/en-us/openjdk/8/html/using_openjdk_8_source-to-image_for_openshift/index"
+  value: &docs "https://access.redhat.com/documentation/en-us/openjdk/8/html/using_openjdk_8_source-to-image_for_openshift/index"
```
jerboaa: Should probably point to the upstream docs too?

ubi8-openjdk-11.yaml
jmtd: OK it's decided; I'll push a subsequent commit that unifies them all with the upstream/community URI. Thank you for the feedback!

ubi8-openjdk-8.yaml
jmtd: Woops, fixed now. Thanks for spotting

jerboaa (approved):

PR #398, opened 2023-09-15
Change GHA runners to ubuntu-latest

jmtd:
This is a hail-mary change: we are experiencing problems with GHA runs timing out (behave tests taking up to five hours to run). We were running on a not-current GHA runner version, and I think the old runners get fewer resources than the newer ones.


jerboaa (approved):
OK if CI looks good.

PR #404, opened 2023-10-09
[OPENJDK-1963] document MAVEN_MIRRORS upcasing behaviour

jmtd:
rewrite description of MAVEN_MIRRORS to properly document the format of the parameter (uppercase, dashes get replaced with underscores).

https://issues.redhat.com/browse/OPENJDK-1963


modules/maven/default/module.yaml
```diff
@@ -31,11 +31,21 @@ envs:
 - name: MAVEN_CLEAR_REPO
   description: If set then the Maven repository is removed after the artifact is built. This is useful for keeping the created application image small, but prevents *incremental* builds. Will be overridden by **S2I_ENABLE_INCREMENTAL_BUILDS**.  Defaults to *false*.
 - name: MAVEN_MIRROR_URL
-  description: The base URL of a mirror used for retrieving artifacts.
+  description: >
+    The base URL of a mirror used for retrieving artifacts.
+    See `MAVEN_MIRRORS` for specifying multiple mirrors.
```
jerboaa: This is already part of #403. Wouldn't this conflict?

modules/maven/default/module.yaml
```diff
@@ -31,11 +31,21 @@ envs:
...
-  example: "dev-one,qe-two"
+  description: >
+    Enables multi-mirror support. Specify a comma-delimited list of capitalized
+    mirror identifiers. The configuration for each mirror will be determined by
+    correspondingly prefixed `MAVEN_MIRROR_*` variables. Any dashes in mirror
+    names will be replaced by underscores. For example: Specifying
+    `DEV-ONE,QE-TWO` configures two mirrors and their URLs will be read from
+    the `DEV_ONE_MAVEN_MIRROR_URL` and `QE_TWO_MAVEN_MIRROR_URL` variables.
+    See also: `prefix_MAVEN_MIRROR_ID`; `prefix_MAVEN_MIRROR_OF`;
+    `prefix_MAVEN_MIRROR_URL`.
```
jerboaa: Not sure what is meant by `prefix_MAVEN_MIRROR_ID`. Is it meant to point to a variable, like if `MAVEN_MIRRORS=one` then `prefix` becomes `ONE_`? If so, how about using `<prefix>_MAVEN_MIRROR_ID` etc. and including the example last, which would tie everything together?

modules/maven/default/module.yaml
jmtd: Yes, whoops. Resolved now, thanks.

modules/maven/default/module.yaml
jmtd: You're guess is right, that's what it's intended to be used as. There's no tooling that puts in cross-reference hyperlinks, (yet), but it's intended to point the reader at the entry for `prefix_MAVEN_MIRROR_URL`, elsewhere. Since `prefix_` is indeed a placeholder, we could adopt a nomenclature to make that clearer. I'll have to experiment to make sure that the GHA scripts don't choke on chevrons.

jmtd: We could use `<prefix>_MAVEN_MIRROR_URL` without breaking anything.

if we used `<<prefix_MAVEN_MIRROR_URL>>`, the doc generator would put an internal hyperlink in to a section labelled with the same name. To make the label, we would need the intermediate asciidoctor for the target text to look like `[[prefix_MAVEN_MIRROR_URL]]prefix_MAVEN_MIRROR_URL`. This could be achieved with 

```yaml
- name: '[[prefix_MAVEN_MIRROR_URL]]prefix_MAVEN_MIRROR_URL'
  description: "The URL of the mirror."
  example: "http://10.0.0.1:8080/repository/internal"
```

but that's ugly. Alternatively the python script could automatically generate the `[[]]` anchor part for every row.

This might be clearer with a POC and I'm conflating two things here

 * adding cross-references between entries
 * replacing `prefix_SOME_THING` to make clearer to humans that `prefix_` is a placeholder

jmtd:
I've pushed a commit that standardizes on `<prefix>_` when referencing an environment variable pattern, and filed https://github.com/jboss-container-images/openjdk/issues/422 to track further improvements to the way we reference variables in descriptions.

jerboaa (approved):

PR #407, opened 2023-10-24
[OPENJDK-2372] singleton-jdk: force-set more java alternatives

jmtd:
This is to fix a race situation where, sometimes, removing JDK8 after installing JDK11 results in the jre_openjdk link group being empty. For further information

* https://issues.redhat.com/browse/OPENJDK-2372
* https://issues.redhat.com/browse/RHEL-3437
* https://issues.redhat.com/browse/RHEL-3444

(UBI8 cherry-pick to follow)

### DIFF
--- a/modules/singleton-jdk/configure.sh
+++ b/modules/singleton-jdk/configure.sh
@@ -15,7 +15,9 @@ fi
 # our stated JAVA_VERSION-JAVA_VENDOR (e.g.: 11-openjdk; 1.8.0-openj9)
 rpm -e --nodeps $(rpm -qa java-* | grep -v "^java-${JAVA_VERSION}-${JAVA_VENDOR}")
 
-# workaround for https://issues.redhat.com/browse/RHEL-3437
+# workaround for <https://issues.redhat.com/browse/RHEL-3437>
+# The alternative link groups touched here need to match up with those set in
+# modules/jdk/*/configure.sh
 _arch="$(uname -m)"
 for alt in java javac java_sdk_openjdk jre_openjdk; do
   alternatives --set "$alt" "java-${JAVA_VERSION}-${JAVA_VENDOR}.${_arch}"
--- a/modules/singleton-jdk/configure.sh
+++ b/modules/singleton-jdk/configure.sh
@@ -16,4 +16,7 @@ fi
 rpm -e --nodeps $(rpm -qa java-* | grep -v "^java-${JAVA_VERSION}-${JAVA_VENDOR}")
 
 # workaround for https://bugzilla.redhat.com/show_bug.cgi?id=2089626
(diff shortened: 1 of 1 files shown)


modules/singleton-jdk/configure.sh
```diff
@@ -16,4 +16,7 @@ fi
 rpm -e --nodeps $(rpm -qa java-* | grep -v "^java-${JAVA_VERSION}-${JAVA_VENDOR}")
 
 # workaround for https://bugzilla.redhat.com/show_bug.cgi?id=2089626
-alternatives --set java "java-${JAVA_VERSION}-${JAVA_VENDOR}.$(uname -m)"
+_arch="$(uname -m)"
```
jerboaa: Any idea why we use `uname -m` here and `uname -i` in the [jdk module configure](https://github.com/jboss-container-images/openjdk/blob/d39eac69f37f4090b89103117fcddbbf24ad680c/modules/jdk/11/configure.sh#L17)? I think we should be consistent throughout.

modules/singleton-jdk/configure.sh
```diff
@@ -16,4 +16,7 @@ fi
 rpm -e --nodeps $(rpm -qa java-* | grep -v "^java-${JAVA_VERSION}-${JAVA_VENDOR}")
 
 # workaround for https://bugzilla.redhat.com/show_bug.cgi?id=2089626
```
jerboaa: Lets reference the migrated JDK bug here: https://issues.redhat.com/browse/RHEL-3437

jerboaa (commented):
We should add a comment referencing the `jdk` module's configure. I.e. mention that these, `java`, `javac`, `java_sdk_openjdk` and `jre_openjdk`, need to get updated in sync.

modules/singleton-jdk/configure.sh
jmtd: I switched to `-m` because, in one context, I was getting `Unknown` with `-i`. I think that  can happen (at least) in a VM. `-m` seems to be more reliable. If you are happy with that, shall I update https://github.com/jboss-container-images/openjdk/blob/d39eac69f37f4090b89103117fcddbbf24ad680c/modules/jdk/11/configure.sh#L17 in this PR?

jerboaa: Yes, please. We need to use the same. Otherwise we ask for more bugs.

jmtd: OK, I think the `uname -i` = `unknown` problem is a Debian thing; so it is probably safer to align *this* instance with the others, and revisit them all in another PR if I hit the problem again. (I couldn't reproduce it just now in a RHEL VM, or with RHEL containers on a Debian host.)

jerboaa (approved):

PR #408, opened 2023-10-24
Openjdk 2384 singleton jdk alternatives

jmtd:
[OPENJDK-2384] singleton-jdk: force-set more java alternatives
    
    This is to fix a race situation where, sometimes, removing JDK8
    after installing JDK11 results in the jre_openjdk link group
    being empty. For further information

* https://issues.redhat.com/browse/OPENJDK-2384
* https://issues.redhat.com/browse/OPENJDK-2372
* https://issues.redhat.com/browse/RHEL-3437
* https://issues.redhat.com/browse/RHEL-3444


jerboaa (commented):
Same comments as in #407 apply here.

jerboaa (approved):

PR #411, opened 2023-11-07
[OPENJDK-2408] s2i: use --archive flag for rsync

jmtd:
Switch from '-rl' to '--archive' which attempts to preserve as much metadata from the source files as possible, including timestamps, the motivation behind OPENJDK-2408.

I haven't written a test yet, I'm just mooting the approach.

### DIFF
--- a/modules/s2i/bash/artifacts/opt/jboss/container/java/s2i/maven-s2i-overrides
+++ b/modules/s2i/bash/artifacts/opt/jboss/container/java/s2i/maven-s2i-overrides
@@ -29,7 +29,7 @@ function maven_s2i_custom_binary_build() {
     binary_dir="${S2I_SOURCE_DIR}"
   fi
   log_info "Copying binaries from ${binary_dir} to ${S2I_TARGET_DEPLOYMENTS_DIR} ..."
-  rsync -rl --out-format='%n' "${binary_dir}"/ "${S2I_TARGET_DEPLOYMENTS_DIR}"
+  rsync --archive --out-format='%n' "${binary_dir}"/ "${S2I_TARGET_DEPLOYMENTS_DIR}"
 }
 
 function maven_s2i_deploy_artifacts_override() {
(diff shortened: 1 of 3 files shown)


jerboaa:
@jmtd All tests seem to timeout (on the default 6h timeout) and get cancelled. Time to bump them?
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

I'd like to see some tests covering the `s2i_core_copy_data` function. It seems that one isn't that much corner-case?

jmtd:
> @jmtd All tests seem to timeout (on the default 6h timeout) and get cancelled. Time to bump them?

Yeah, I've started investigating why they time out (#410 relates to that) and they seem to hang when behave-test-steps runs "execute" for a running command via docker-py. But I've no clue why they hang. If by bump you mean shortening the timeout from the GHA side, that's definitely worth doing yes. I was going to try the same from the behave-test-steps side. I'll take a look at that

jmtd:
I just ran all the tests for this branch and ubi9/openjdk-17 and they passed.
For the continuing GHA problems, I'm keeping a narrative in #418.

jerboaa:
> I'd like to see some tests covering the `s2i_core_copy_data` function. It seems that one isn't that much corner-case?

@jmtd Have you considered that?

jmtd:
> @jmtd Have you considered that?

No sorry, I missed your original comment. Let's make sure we have coverage before we merge.

What I was pondering whas how to construct a test situation that had timestamps in it, since git doesn't record timestamps. I'm sure we can work around that with an appropriate hook somewhere.

jerboaa:
> > @jmtd Have you considered that?
> 
> No sorry, I missed your original comment. Let's make sure we have coverage before we merge.
> 
> What I was pondering whas how to construct a test situation that had timestamps in it, since git doesn't record timestamps. I'm sure we can work around that with an appropriate hook somewhere.

I'm not sure we need timestamps preserved if a test artefact comes from git. What we do want to preserve is the timestamp (modified date) from the time of checkout `T` to the one ending up in the final image `T'`. `T'` should be identical to `T`. An AppCDS test comes to mind (perhaps with a binary artefact). AFAIK, `rsync` would be in between when `T` and `T'` happens. Assuming the build isn't instant (`0 secs`), those timestamps might differ.

jmtd:
The new test here references my fork of openjdk-test-applications (PR for that fork is https://github.com/jboss-container-images/openjdk-test-applications/pull/4). It fails on current images and passes after the other commit in this PR.

jerboaa (approved):
Looks good. Other than the reference to the fork. Please change before you push. Thanks for adding the test!

PR #414, opened 2023-11-08
[OPENJDK-2410] Add UBI9 JDK21 images

jmtd:
https://issues.redhat.com/browse/OPENJDK-2410


redhat/ubi9-openjdk-21-runtime.yaml
```diff
@@ -0,0 +1,27 @@
+osbs:
+  configuration:
+    container:
+      compose:
+        pulp_repos: true
+        packages:
+        - java-21-openjdk-headless
+        signing_intent: release
```
jerboaa, line 8: Should we add architectures here?

```
      # Currently only on ppc64le, aarch64, x86_64
      platforms:
        only:
          - x86_64
          - aarch64
          - ppc64le
```

jerboaa (approved):
Seems OK. Consider adding a restrict set of arches.

PR #415, opened 2023-11-09
[OPENJDK-2411] JDK21 images for UBI8

jmtd:
https://issues.redhat.com/browse/OPENJDK-2411


jmtd:
I am doing a round of scratch builds for these and running the behave test suite locally.

jmtd:
> I am doing a round of scratch builds for these and running the behave test suite locally.

all passed

ubi8-openjdk-21.yaml
```diff
@@ -0,0 +1,56 @@
...
+- name: "io.k8s.display-name"
+  value: "Java Applications"
+- name: "io.openshift.tags"
+  value: "builder,java"
+- name: "maintainer"
+  value: "Red Hat OpenJDK <openjdk@redhat.com>"
+- name: "com.redhat.component"
+  value: "openjdk-21-ubi8-container"
+- name: "usage"
+  value: "https://access.redhat.com/documentation/en-us/red_hat_jboss_middleware_for_openshift/3/html/red_hat_java_s2i_for_openshift/"
```
jerboaa: Should this point to?

https://jboss-container-images.github.io/openjdk/

ubi8-openjdk-21.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+modules:
+  repositories:
+  - path: modules
+  install:
+  - name: jboss.container.util.pkg-update
+  - name: jboss.container.openjdk.jdk
+    version: "21"
+  - name: jboss.container.maven
+    version: "3.8.17"
```
jerboaa, line 50: `3.8.21`?

jerboaa (approved):

ubi8-openjdk-21.yaml
jmtd: Good catch. At the moment there is no `maven-openjdk21` RPM package, which would be what `3.8.21` should install (once the module exists), but we expect there to be in the future. So I think for now it's ok to use the 17 one in conjunction with singleton-jdk.

ubi8-openjdk-21.yaml
jmtd: Thank you yes: resolved in 4eff5d8d6293b84de7221c2b8551e122390fcf0b.

ubi8-openjdk-21.yaml
jerboaa: OK.

sefroberg: @jmtd is there a ticket to track the creation of maven-openjdk21?

jmtd: There is for RHEL 9.4 (RHEL-13046), but not (yet) for 8.10. I asked the filer of the former whether they'd like me to clone for the latter and haven't yet had a response.

PR #418, opened 2023-11-14
Fix GHA by various methods

jmtd:
Get the GHA CI lights green again.

 * For some reason the GHA environment is configured with some docker.io credentials (Perhaps this is to do with interacting with ghcr.io). [s2i seems to query remote repositories even when it's not planning to pull an image](https://github.com/openshift/source-to-image/issues/1134). Explicitly removing these (via logging out) works around this
 * Various tests, for still-unknown reasons, freeze indefinitely when run in GHA. Referencing my fork of behave-test-steps makes use of [my changes to "watchdog" some docker API calls via a sub-process and abort if it does not complete within a timeout period](https://github.com/cekit/behave-test-steps/pull/50). (cekit always fetches the v1 branch of behave-test-steps, at least [until --steps-ref ships](https://github.com/cekit/cekit/pull/877)). This changes those tests from freezing the whole run, to just failing those tests.
 * Some label tests fail as I updated them in a recent PR and didn't update the images to match. Add name and version labels to all images to get that one passing.
 * Finally, tag most of the s2i/maven/http proxy tests, as well as a "java perf dir owner" test "ignore", which means they're skipped entirely. These tests are still failing, but I would prefer to get the CI functioning again for all the other tests whilst I continue to investigate them.

### DIFF
--- a/.github/workflows/image-workflow-template.yml
+++ b/.github/workflows/image-workflow-template.yml
@@ -42,4 +42,4 @@ jobs:
       - name: Behave Tests
         run: |
           echo /home/runner/work/_temp/openshift-bin >> $GITHUB_PATH
-          cekit -v --descriptor ${{ inputs.image }}.yaml test behave
+          cekit -v --descriptor ${{ inputs.image }}.yaml test behave --steps-url https://github.com/jmtd/behave-test-steps
--- a/.github/workflows/image-workflow-template.yml
+++ b/.github/workflows/image-workflow-template.yml
@@ -33,6 +33,12 @@ jobs:
           github_pat: ${{ github.token }}
           s2i: "1.3.4"
 
+      # s2i misbehaves if registry credentials are present: tries and fails to query
+      # image metadata from docker.io before each build, etc. See:
+      # https://github.com/openshift/source-to-image/issues/1134
+      - name: clear docker credentials
+        run: docker logout
+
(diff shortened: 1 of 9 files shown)


jmtd:
Doesn't seem to have helped. Might try walking back cekit versions

jmtd:
The Docker Python library's client object has a timeout property (default = 60 seconds) which is supposed to be applied to all the API calls. However it is not applied to the `exec_start` call ( or at least, with the library version and API version that is in use here):

```python
>>> d
<docker.api.client.APIClient object at 0x7f95d62f6250>
>>> d.timeout
60
>>> d.start(container="1638a53efc01")
>>> inst = d.exec_create(container="1638a53efc01", cmd=["bash","-c","echo before;sleep 120;echo after"])
>>> d.exec_start(inst, detach=False)
b'before\nafter\n'
```

IF the timeout was active, the command should have returned prior to the "sleep" completing, and we would not see the output of the remainder of the shell command.

THis is docker-6.1.3.dist-info,  'ApiVersion': '1.41' (default for this library version), [although behave-test-steps hard-codes 1.22](https://github.com/cekit/behave-test-steps/blob/v1/steps/container.py#L37C1-L37C41).

jerboaa (approved):

jerboaa:
Nice work!

PR #424, opened 2023-12-05
[OPENJDK-2519] Add tzdata module to reinstall tzdata

jmtd:
tzdata is deliberately broken in ubi-minimal images (presumably for disk space reasons). The fix is to reinstall the tzdata package:

    <https://access.redhat.com/solutions/5616681>

This appears to add a neglibible amount to our image sizes (<1%) but resolves TZ issues for Java applications.

https://issues.redhat.com/browse/OPENJDK-2519

### DIFF
--- a/modules/util/tzdata/execute.sh
+++ b/modules/util/tzdata/execute.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+set -euo pipefail
+
+for candidate in yum dnf microdnf; do
+    if command -v "$candidate"; then
+        mgr="$(command -v "$candidate")"
+        "$mgr" reinstall tzdata -y && \
+            "$mgr" -y clean all
+        exit
+    fi
+done
+
+echo "cannot find a package manager" >&2
+exit 1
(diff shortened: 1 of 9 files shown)


modules/util/tzdata/execute.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -euo pipefail
+
+for candidate in yum dnf microdnf; do
+    if command -v "$candidate"; then
+        mgr="$(command -v "$candidate")"
+        "$mgr" reinstall tzdata -y && \
```
jerboaa, line 7: Should this re-install `tzdata` and `tzdata-java`?

jmtd: `tzdata-java` appears to be intact (`rpm -V` is quiet). Unless there's something crucial in one of its scriptlets relying on `tzdata` (I'll take a look), there's no need.

jerboaa: OK. Thanks for checking.

jerboaa (approved):

modules/util/tzdata/execute.sh
jmtd: no scriptlets.

PR #427, opened 2023-12-18
[Openjdk-2587] reinstall tzdata rpm for  ubi8

jmtd:
https://issues.redhat.com/browse/OPENJDK-2587

See also #424

### DIFF
--- a/modules/util/tzdata/execute.sh
+++ b/modules/util/tzdata/execute.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+set -euo pipefail
+
+for candidate in yum dnf microdnf; do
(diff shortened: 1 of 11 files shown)


jerboaa:
Are we missing some test fixes?

jmtd:
I better check out these GHA failures

jmtd:
Last remaining failure is a build (not test) one, and transient relating to tzdata. It's possible we will have to make the tzdata reinstall more robust (if the installed-but-broken version differs from that available on the mirrors I think that provokes a problem) but I'm not sure we have enough data on how common this problem is yet.

jerboaa (approved):

PR #429, opened 2023-12-21
[OPENJDK-2177] Move behave tests to relevant modules

jhuttana:
JIRA ticket : https://issues.redhat.com/browse/OPENJDK-2177

The approved tests I have moved to relevant modules.
Tested locally after moving and they worked fine .
NOTE: To quickly check only the moved tests , I marked them as @wip while testing.
```
$ cekit --descriptor ubi9-openjdk-17.yaml test behave --wip
.
.
.

  @ubi9 @wip
  Scenario: Check the attributes of /home/default using stat              # features/jboss.container.user/general.feature:5
    When container is started with args                                   # steps/container_steps.py:86 10.503s
      | arg     | value              |
      | command | stat /home/default |
    Then available container log should contain Access: (0770/drwxrwx---) # steps/container_steps.py:50
    Then available container log should contain Access: (0770/drwxrwx---) # steps/container_steps.py:50 0.004she logs
2023-12-22 00:53:57,897 container.py:100        INFO  Removing container '308c3900d5357c99cf08de4d903b58510292b96ca9c61db450ba089f0d79c7e9', 1 try...
2023-12-22 00:53:57,921 container.py:103        INFO  Container '308c3900d5357c99cf08de4d903b58510292b96ca9c61db450ba089f0d79c7e9' removed

6 features passed, 0 failed, 6 skipped
19 scenarios passed, 0 failed, 46 skipped
49 steps passed, 0 failed, 139 skipped, 0 undefined
Took 14m8.794s
2023-12-22 00:53:57,921 cli.py:503        INFO  Finished!
```

@jmtd Could you please take a look?


jmtd (approved):

jmtd:
I'm happy to merge this, but I'd like @jfdenise to take a quick look first please.
JF, I think there's a chance that this change will cause some of our tests to potentially be in scope for the wildfly images that reference this repository as a module source. Would you be OK to take a look and see what happens based on this branch?

jfdenise:
@jmtd , you are right, it is going to create an issue, the tests are brought in when testing wildfly-s2i image. I looked at the test and they can't be run with wildfly-s2i image. Even if we are not currently using UBI9, we will at some point.

jmtd:
> @jmtd , you are right, it is going to create an issue, the tests are brought in when testing wildfly-s2i image. I looked at the test and they can't be run with wildfly-s2i image. Even if we are not currently using UBI9, we will at some point.

So Cekit collects them, OK. Does our test tagging clash with yours? Perhaps they aren't tagged correctly to be executed by you?

jfdenise:
@jmtd , I forgot about the tagging filtering. Perhaps make your tags to fully contain the image (not just uib9 as I can see in some of them that target both runtime and builder). I don't suspect that we use ubi9/xxx but it seems safer. So in the end, even if the tests are brought in they should be ignored by our test execution.

jmtd:
Hi @jfdenise OK, so, at the moment you are only using our ubi8 branch, so this PR in particular won't impact you at all; but we will backport this change to the ubi8 branch too.

After that, your naming scheme and ours are different enough that the tags won't match up unless one or both of us change our naming scheme. I think we should look to expand our short tags (ubi8 etc) only if that is going to happen.

Finally, you might actually *want* to run some of these tests: the intention of moving them is to get the tests closer to the code they are testing, so, if you are using a particular module, you might benefit from running its tests. However many won't work with wildfly (any of our S2I ones I guess). For this situation, if you wanted to enable a test, we would happily commit additional, wildfly-specific tags to the behave files.

Thank you for taking a look!

PR #441, opened 2024-01-30
Jlink dockerfile improvements

jmtd:
I made some improvements to the jlink Dockerfile. These might be UBI9 specific (I haven't retested with UBI8). Resulting images are a bit smaller and crucially run with a non-root UID.


jhuttana (approved):
LGTM

PR #443, opened 2024-02-13
[Openjdk-2587] ensure pkg-update happens before reinstall tzdata (ubi8)

jmtd:
https://issues.redhat.com/browse/OPENJDK-2587

The tzdata module, which calls "microdnf reinstall tzdata", may fail if the base image has an installed tzdata RPM that is no longer available on the mirrors / is out-of-date. For example

    base image installed RPM: tzdata-2023d-1.el8.noarch
    attempt to build ubi8-openjdk-17 (where tzdata module is first)
    "Installed package tzdata-2023d-1.el8.noarch not available"

The fix is to move the pkg-update module to be earlier. Do this for all images, for consistency, place pkg-update first. (This matches the ubi9 images)

I have not included a unit test as I am not sure how to approach mocking older/newer RPM versions in a controllable way.

### DIFF
--- a/modules/util/tzdata/module.yaml
+++ b/modules/util/tzdata/module.yaml
@@ -1,7 +1,14 @@
 schema_version: 1
 name: jboss.container.util.tzdata
 version: '1.0'
-description: Reinstall the tzdata package, to ensure zoneinfo is present
+description: Reinstall the tzdata package, to ensure zoneinfo is present.
+
+# if the base image tzdata version is not available on the RPM mirrors (such as
+# when it has been updated), the reinstall action will fail. To prevent this,
+# run pkg-update first.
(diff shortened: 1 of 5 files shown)


jerboaa (approved):
OK. A few comments would probably be nice.

jmtd:
> OK. A few comments would probably be nice.

Fair enough. I decided to write a comment to explain, and that led me to realise that, we can (and probably should) have a module dependency between tzdata and pkg-update, so I added that. It's a no-op on top of the other change; but it would catch this in future if we change up the modules in the image again. If you ACK this I'll forward-port the last commit back to UBI9 in another PR.

jerboaa (approved):

PR #449, opened 2024-02-22
OPENJDK-2736 and OPENJDK-2737: gate Jlink code, fail early if jmods unavailable

jmtd:
* https://issues.redhat.com/browse/OPENJDK-2736
 * https://issues.redhat.com/browse/OPENJDK-2737

### DIFF
--- a/modules/jlink/artifacts/opt/jboss/container/java/jlink/preflight.sh
+++ b/modules/jlink/artifacts/opt/jboss/container/java/jlink/preflight.sh
@@ -4,9 +4,11 @@ jlink_preflight_check()
     if [ "$JAVA_VERSION" -lt 11 ]; then
       echo "Jlink integration not available for JDK${JAVA_VERSION}!"
       echo "Jlink integration is only supported for JDK versions 11 and newer."
+      exit 1
     fi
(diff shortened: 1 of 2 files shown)


modules/jlink/artifacts/opt/jboss/container/java/jlink/preflight.sh
```diff
@@ -0,0 +1,12 @@
+jlink_preflight_check()
+{
+    # preflight check: do we have what we need?
+    if [ "$JAVA_VERSION" -lt 11 ]; then
+      echo "Jlink integration not available for JDK${JAVA_VERSION}!"
+      echo "Jlink integration is only supported for JDK versions 11 and newer."
```
jhuttana, line 6: I feel if the JAVA_VERSION is less than 11 then we can stop doing the further steps and exit here itself with the above message. WDYT?

jmtd: Yes; there should be an `exit 1` here, and also in the fail branch below. Thanks!

Josh-Matsuoka:
Looks good to me once Jaya's comment is incorporated

PR #450, opened 2024-02-26
[OPENJDK-2731] Behave test for jlink integration code

jhuttana:
Could you please review ?

### DIFF
--- a/modules/jlink/tests/features/jlink.feature
+++ b/modules/jlink/tests/features/jlink.feature
@@ -1,0 +1,16 @@
+@ubi9/openjdk-11
+@ubi9/openjdk-17
+@ubi9/openjdk-21
+Feature: Openshift OpenJDK S2I tests (jlink specific)
+Scenario: Ensure jlinked builder is used to build the containerized application image
+      Given s2i build https://github.com/jboss-container-images/openjdk-test-applications from quarkus-quickstarts/getting-started-3.0.1.Final-nos2i
+       | variable            | value        |
+       | S2I_ENABLE_JLINK    | true         |
+       | QUARKUS_PACKAGE_TYPE| uber-jar     |
+      Then run ls /tmp/jre in container and check its output for bin
+
+Scenario: Ensure S2I_ENABLE_JLINK=false is the default
+      Given container is started with env
+       | variable            | value        |
+       | QUARKUS_PACKAGE_TYPE| uber-jar     |
+      Then container log should not contain S2I_ENABLE_JLINK=true
--- a/modules/jlink/tests/features/jlink.feature
+++ b/modules/jlink/tests/features/jlink.feature
@@ -12,4 +12,4 @@ Scenario: Ensure jlinked builder is used to build the containerized application 
 Scenario: Ensure S2I_ENABLE_JLINK is not set to true
       Given s2i build https://github.com/jboss-container-images/openjdk-test-applications from quarkus-quickstarts/getting-started-3.0.1.Final-nos2i
       Then container log should not contain S2I_ENABLE_JLINK=true
-      And run ls /tmp/jre in container and check its output for No such file or directory
+      And run bash -c "test ! -d /tmp/jre && echo PASS" in container and immediately check its output for PASS
--- a/modules/jlink/tests/features/jlink.feature
+++ b/modules/jlink/tests/features/jlink.feature
@@ -9,6 +9,7 @@ Scenario: Ensure jlinked builder is used to build the containerized application 
        | QUARKUS_PACKAGE_TYPE| uber-jar     |
       Then run ls /tmp/jre in container and check its output for bin
 
-Scenario: Ensure S2I_ENABLE_JLINK=false is the default
-      Given container is started with command env
+Scenario: Ensure S2I_ENABLE_JLINK is not set to true
+      Given s2i build https://github.com/jboss-container-images/openjdk-test-applications from quarkus-quickstarts/getting-started-3.0.1.Final-nos2i
       Then container log should not contain S2I_ENABLE_JLINK=true
+      And run ls /tmp/jre in container and check its output for ls:
--- a/modules/jlink/tests/features/jlink.feature
+++ b/modules/jlink/tests/features/jlink.feature
@@ -10,7 +10,5 @@ Scenario: Ensure jlinked builder is used to build the containerized application 
       Then run ls /tmp/jre in container and check its output for bin
 
 Scenario: Ensure S2I_ENABLE_JLINK=false is the default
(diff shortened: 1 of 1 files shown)


modules/jlink/tests/features/jlink.feature
```diff
@@ -0,0 +1,16 @@
...
+Feature: Openshift OpenJDK S2I tests (jlink specific)
+Scenario: Ensure jlinked builder is used to build the containerized application image
+      Given s2i build https://github.com/jboss-container-images/openjdk-test-applications from quarkus-quickstarts/getting-started-3.0.1.Final-nos2i
+       | variable            | value        |
+       | S2I_ENABLE_JLINK    | true         |
+       | QUARKUS_PACKAGE_TYPE| uber-jar     |
+      Then run ls /tmp/jre in container and check its output for bin
+
+Scenario: Ensure S2I_ENABLE_JLINK=false is the default
+      Given container is started with env
```
jmtd: this scenario won't need an environment table

jmtd: for this test to work the container needs to run the command `env`, see [this step](https://github.com/cekit/behave-test-steps/blob/v1/steps/container_steps.py#L106)

jhuttana:
I have pushed my recent changes to address the review comments but the latest commit is not showing up in this PR!!
Whereas I can see the commit on my fork-branch : https://github.com/jhuttana/openjdk/commits/jlinkcode_integration/

jhuttana:
> I have pushed my recent changes to address the review comments but the latest commit is not showing up in this PR!! Whereas I can see the commit on my fork-branch : https://github.com/jhuttana/openjdk/commits/jlinkcode_integration/

Meanwhile I was trying some hacks like changing the base etc and then the latest commit appeared :D Not sure what worked but its fine now :)

jmtd:
Hi Jaya, I've been thinking about `S2I_ENABLE_JLINK` and I think we should alter it so that it is NOT defined by default (to true or false) and so meets Cekit's definition of a "configuration environment variable". I've filed [OPENJDK-2806](https://issues.redhat.com/browse/OPENJDK-2806) (S2I_ENABLE_JLINK should be a configuration env var only) and PR #453, but it  means changes to this PR too.

I think the second test is OK, but should be renamed to describe that it is not testing for `S2I_ENABLE_JLINK=false`, but that it's _not_ `S2I_ENABLE_JLINK=true`;

The first test is OK, but for completion can we also have another which performs the inverse? I.e., when `S2I_ENABLE_JLINK` is _not_ set by the user, test that `/tmp/jre` does not exist.

jmtd:
Please also add a test for the JDK8 image that proves that `S2I_ENABLE_JLINK=true` does not work for that image; My changes in #449 should make this a fatal error so we might need to experiment with the test to figure out how to catch that behaviour.

jhuttana:
> Hi Jaya, I've been thinking about `S2I_ENABLE_JLINK` and I think we should alter it so that it is NOT defined by default (to true or false) and so meets Cekit's definition of a "configuration environment variable". I've filed [OPENJDK-2806](https://issues.redhat.com/browse/OPENJDK-2806) (S2I_ENABLE_JLINK should be a configuration env var only) and PR #453, but it means changes to this PR too.
> 
> I think the second test is OK, but should be renamed to describe that it is not testing for `S2I_ENABLE_JLINK=false`, but that it's _not_ `S2I_ENABLE_JLINK=true`;

+1  makes more sense 

> 
> The first test is OK, but for completion can we also have another which performs the inverse? I.e., when `S2I_ENABLE_JLINK` is _not_ set by the user, test that `/tmp/jre` does not exist.

If my understanding is right then, logically  `S2I_ENABLE_JLINK` is not set and `S2I_ENABLE_JLINK=false` are both same. 
In both  cases `/tmp/jre`  is not created as it should not take the jlink path. So if at all we want to test `/tmp/jre` doesn't exist then we can club it with second Scenario. WDYT?

modules/jlink/tests/features/jlink.feature
```diff
@@ -0,0 +1,15 @@
...
+      Given s2i build https://github.com/jboss-container-images/openjdk-test-applications from quarkus-quickstarts/getting-started-3.0.1.Final-nos2i
+       | variable            | value        |
+       | S2I_ENABLE_JLINK    | true         |
+       | QUARKUS_PACKAGE_TYPE| uber-jar     |
+      Then run ls /tmp/jre in container and check its output for bin
+
+Scenario: Ensure S2I_ENABLE_JLINK is not set to true
+      Given s2i build https://github.com/jboss-container-images/openjdk-test-applications from quarkus-quickstarts/getting-started-3.0.1.Final-nos2i
+      Then container log should not contain S2I_ENABLE_JLINK=true
+      And run ls /tmp/jre in container and check its output for ls:
```
jmtd: I guess `ls:` here is a string that ls prints when it doesn't find something, but it's not clear. Could we look for `No such file or directory` instead?

jmtd:
>  In both cases /tmp/jre is not created as it should not take the jlink path. So if at all we want to test /tmp/jre doesn't exist then we can club it with second Scenario. WDYT?

Yes that makes sense.

modules/jlink/tests/features/jlink.feature
```diff
@@ -0,0 +1,15 @@
...
+      Given s2i build https://github.com/jboss-container-images/openjdk-test-applications from quarkus-quickstarts/getting-started-3.0.1.Final-nos2i
+       | variable            | value        |
+       | S2I_ENABLE_JLINK    | true         |
+       | QUARKUS_PACKAGE_TYPE| uber-jar     |
+      Then run ls /tmp/jre in container and check its output for bin
+
+Scenario: Ensure S2I_ENABLE_JLINK is not set to true
+      Given s2i build https://github.com/jboss-container-images/openjdk-test-applications from quarkus-quickstarts/getting-started-3.0.1.Final-nos2i
+      Then container log should not contain S2I_ENABLE_JLINK=true
+      And run ls /tmp/jre in container and check its output for No such file or directory
```
jmtd: I'm not able to get this test to pass when it should locally. I'm not sure why: stderr output not in container log? case sensitivity? locale sensitivity for the format of the ls error output? I see this:
`      Exception: ("Phrase 'No such file or directory' was not found in the output of running the 'ls /tmp/jre' command", b"ls: cannot access '/tmp/jre': No such file or directory\n", ExecException('Command ls /tmp/jre failed to execute, return code: 2'))
`

jmtd: I've come up with an alternative solution, I'll push it

jhuttana:
Thank you :)

PR #453, opened 2024-02-27
[OPENJDK-2806] Remove value for S2I_ENABLE_JLINK

jmtd:
S2I_ENABLE_JLINK is a configuration environment variable, not an informational one, and so the default should be unset, rather than "false".

https://issues.redhat.com/browse/OPENJDK-2806


jmtd:
Runtime images tests are all failing, I think due to #455.

jhuttana:
LGTM

jhuttana (approved):

PR #454, opened 2024-02-28
[OPENJDK-2686] Jlink stage 1 - BuildConfig and Imagestream for creating a builder image with the jmods rpm installed

Josh-Matsuoka:
Addresses stage 1 of the planned multi-stage jlink build

-  Adds a script and set of 3 buildconfigs under templates/jlink/jlink-builder, one each for JDK11, JDK17, and JDK21
- These builds that these buildconfigs produce named jlink-builder-jdk-$VERSION
- Running the script creates an ImageStream named ubi9-openjdk-$VERSION-jlink that the buildconfig will push to when the build is run
- After this ImageStream is created the buildconfig is passed in to openshift and a build is run off of it, pushing the result to the ubi9-openjdk-$VERSION-jlink ImageStream for consumption by stage 2.

The script currently works off of a local crc instance (getting oc from crc oc-env), this can probably be generalized later if needed but for local testing it should do the trick. Requires you to be logged into the crc environment (i,e, having run oc login)


jmtd:
Runtime GHA test failures are likely unrelated to this PR and due to the issue under #455.

templates/jlink/jlink-builder/jdk-11-buildconfig.yaml
```diff
@@ -0,0 +1,23 @@
...
+spec:
+  source:
+    type: Git
+    git:
+      uri: https://github.com/jmtd/quarkus-quickstarts.git
+    contextDir: getting-started
+    dockerfile: |
+      FROM registry.access.redhat.com/ubi9/openjdk-11 AS ubi9-openjdk-11-jmods
+      USER 0
+      RUN microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y java-11-openjdk-jmods
```
jmtd: Please add `USER 185` to switch the runtime user back to non-root at the end

templates/jlink/jlink-builder/build-imagestreams.sh
```diff
@@ -0,0 +1,9 @@
+eval $(crc oc-env)
+VERSION=11
+
+# Buildconfigs attempt to push to ubi9-openjdk-$VERSION-jmods, this needs to exist first
+oc create imagestream ubi9-openjdk-$VERSION-jlink
```
jmtd: Ultimately I hope we don't need a wrapper shellscript, but it is convenient during development.

We could get rid of this line if we include the ImageStream definition in the YAMLs, below. try `oc get -f yaml is/ubi9-openjdk-11-jlink` and then pre-pend the result to `jdk-11-buildconfig.yaml`, separated by `---`.

templates/jlink/jlink-builder/jdk-11-buildconfig.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: build.openshift.io/v1
+kind: BuildConfig
+metadata: 
+  name: jlink-builder-jdk-11
```
jmtd: The JDK version will probably eventually be a template parameter for an OpenShift template. If keeping the three buildconfig YAMLs in sync proves to be a burden, consider deleting two of them (assuming there's no substantive difference besides JDK version) and we can re-introduce them once the content stabilises. Your call!

templates/jlink/jlink-builder/jdk-11-buildconfig.yaml
```diff
@@ -0,0 +1,23 @@
...
+  labels:
+    app: jlink-builder-jdk-11
+spec:
+  source:
+    type: Git
```
jmtd: keys type, git, git.uri and contextDir should be deleted;

templates/jlink/jlink-builder/jdk-11-buildconfig.yaml
```diff
@@ -0,0 +1,23 @@
...
+    git:
+      uri: https://github.com/jmtd/quarkus-quickstarts.git
+    contextDir: getting-started
+    dockerfile: |
+      FROM registry.access.redhat.com/ubi9/openjdk-11 AS ubi9-openjdk-11-jmods
```
jmtd: It should be possible to set `FROM -`, and for OpenShift to supply the correct base image URI. But we would also need to add some more metadata to the buildConfig (see another comment)

jmtd:
I think we should add a dependency on an OpenJDK imageStream, the problem is we don't provide any UBI9 ones yet. We provide

* "ubi8-openjdk-8"
* "ubi8-openjdk-11"
* "ubi8-openjdk-17"
* "ubi8-openjdk-21"
* "java"

I've filed https://issues.redhat.com/browse/OPENJDK-2811 to get some added. In the meantime we should probably add a rudimentary one to this branch; I've filed https://issues.redhat.com/browse/OPENJDK-2812 for that.

templates/jlink/jlink-builder/jdk-11-buildconfig.yaml
jmtd: I've added a basic ubi9 imagestream template in https://github.com/jboss-container-images/openjdk/pull/456,  please take a look, then we can make the suggested change here

templates/jlink/jlink-builder/jdk-17-buildconfig.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: build.openshift.io/v1
+kind: BuildConfig
+metadata: 
+  name: jlink-builder-jdk-17
+  labels:
+    app: jlink-builder-jdk-17
+spec:
+  source:
+    dockerfile: |
+      FROM registry.access.redhat.com/ubi9/openjdk-17 AS ubi9-openjdk-17-jmods
```
jmtd: change to `FROM -` , to make explicit that we're going to rely on OpenShift substituting the FROM image with a local reference to an ImageStreamTag. See: https://docs.openshift.com/container-platform/4.15/cicd/builds/build-strategies.html#builds-strategy-docker-from-image_build-strategies for the additional keys we need to add to make that work (`strategy.dockerStrategy.from.kind` and `strategy.dockerStrategy.from.name`, pointing at the UBI9 ImageStream we merged)

templates/jlink/jlink-builder/jlink-builder-template.yaml
```diff
@@ -0,0 +1,48 @@
...
+- apiVersion: image.openshift.io/v1
+  kind: ImageStream
+  metadata:
+    generation: 1
+    name: ubi9-openjdk-${JDK_VERSION}-jlink
+    namespace: default
+  spec:
+    lookupPolicy:
+      local: false
+  status:
```
jmtd: we can delete `status`  and everything under it, and openshift will populate them at creation time

templates/jlink/jlink-builder/jlink-builder-template.yaml
```diff
@@ -0,0 +1,48 @@
...
+    description: Template to produce an imagestream and buildconfig for a Jlink builder image
+  name: jlink-builder-template
+  namespace: openshift
+objects:
+- apiVersion: image.openshift.io/v1
+  kind: ImageStream
+  metadata:
+    generation: 1
+    name: ubi9-openjdk-${JDK_VERSION}-jlink
+    namespace: default
```
jmtd: delete `namespace: default`, openshift will provide the value at creation time

templates/jlink/jlink-builder/jlink-builder-template.yaml
```diff
@@ -0,0 +1,48 @@
...
+metadata:
+  annotations:
+    description: Template to produce an imagestream and buildconfig for a Jlink builder image
+  name: jlink-builder-template
+  namespace: openshift
+objects:
+- apiVersion: image.openshift.io/v1
+  kind: ImageStream
+  metadata:
+    generation: 1
```
jmtd: can remove `generation: 1`

templates/jlink/jlink-builder/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Stage 1: create ubi9 jlink imagestream and push a ubi9 image with jmods installed to it.
+
+You need:
+
+1. Access to an OpenShift instance, such as crc
+
+Steps to produce the imagestream and image:
+
+1. Switch to the openshift project and install the template
```
jmtd: pre-step needed: `oc create -f templates/ubi9-community-image-streams.json`

templates/jlink/jlink-builder/jlink-builder-template.yaml
```diff
@@ -0,0 +1,48 @@
+---
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  annotations:
+    description: Template to produce an imagestream and buildconfig for a Jlink builder image
+  name: jlink-builder-template
+  namespace: openshift
```
jmtd: delete `namespace: openshift`

jmtd:
This is awesome!

templates/jlink/jlink-builder/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+oc project openshift
```
jmtd, line 15: Indent all the example input/output lines by 4 or more spaces and they'll be rendered in monotype

templates/jlink/jlink-builder/README.md
```diff
@@ -0,0 +1,56 @@
...
+The template for now defines a single parameter, JDK_VERSION. Setting this will set the version of the builder image used in the BuildConfig. Currently suppoted values are 11, 17, and 21. By default this will be 11.
+
+oc process --parameters jlink-builder-template
+
+NAME                DESCRIPTION                                GENERATOR           VALUE
+JDK_VERSION         JDK version to produce a jmods image for                       11
+
+In order to set the JDK version, you will need to use the -p flag of oc process. To process the template and create the imagestreams, simply run 
+
+oc process -n openshift jlink-builder-template -p JDK_VERSION=11 | oc create -f -
```
jmtd: remove `-n openshift` here

templates/jlink/jlink-builder/README.md
jmtd: remove `switch to the openshift project and`: It's normal to use that namespace but not essential. OOTB with CRC for example the developer user cannot write to it. I'm following along here entirely in a separate project.

templates/jlink/jlink-builder/README.md
jmtd: remove this line

templates/jlink/jlink-builder/README.md
```diff
@@ -0,0 +1,56 @@
...
+You need:
+
+1. Access to an OpenShift instance, such as crc
+
+Steps to produce the imagestream and image:
+
+1. Switch to the openshift project and install the template
+
+oc project openshift
+oc create -f templates/jlink/jlink-builder/jlink-builder-template.yaml
```
jmtd: Indent all the example input/output lines by 4 or more spaces and they'll be rendered in monotype

PR #465, opened 2024-03-15
[OPENJDK-2722] Split the tests inside tests/features/java/runtime.feature to couple them with appropriate modules

jhuttana:

### DIFF
--- a/modules/jvm/api/tests/features/runtime.feature
+++ b/modules/jvm/api/tests/features/runtime.feature
@@ -5,13 +5,6 @@
 Feature: OpenJDK Runtime tests 
 
   @ubi8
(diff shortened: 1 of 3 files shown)


jmtd:
The test 'Check JAVA_APP_NAME can contain spaces (OPENJDK-104)' now in modules/jvm/api/tests/features/runtime.feature should probably be in modules/run instead: all the handling of `JAVA_APP_NAME` is in there. Same with the tests for `JAVA_APP_DIR`.

Similarly, `JAVA_OPTIONS` is only referenced in modules/s2i/bash so that test should be there.

jmtd (approved):

PR #470, opened 2024-03-27
[OPENJDK-2850] assemble: binary: Don't set times on directories

jmtd:
If the S2I build (in `--as-dockerfile` mode, for the standalone s2i, or by default inside OpenShift) runs as a user other than 185 (such as s2i's default of 1001), attempting to set the timestamp of /deployments will fail, causing the build to fail.

https://issues.redhat.com/browse/OPENJDK-2850

See also

- [x] test coverage
- [x] need confirmation that this does not regress OPENJDK-2408

### DIFF
--- a/modules/s2i/bash/artifacts/opt/jboss/container/java/s2i/maven-s2i-overrides
+++ b/modules/s2i/bash/artifacts/opt/jboss/container/java/s2i/maven-s2i-overrides
@@ -29,7 +29,12 @@ function maven_s2i_custom_binary_build() {
     binary_dir="${S2I_SOURCE_DIR}"
   fi
   log_info "Copying binaries from ${binary_dir} to ${S2I_TARGET_DEPLOYMENTS_DIR} ..."
-  rsync --archive --out-format='%n' "${binary_dir}"/ "${S2I_TARGET_DEPLOYMENTS_DIR}"
+
+  ( # OPENJDK-2850: use glob (dotglob to match hidden files) to stop rsync altering
+    # timestamps of S2I_TARGET_DEPLOYMENTS_DIR. Don't alter parent shell's dotglob.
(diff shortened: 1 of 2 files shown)


jmtd:
Test "  Scenario: Ensure mtime is preserved for build artifacts (OPENJDK-2408)" passes

jmtd:
New test passes locally (test suite on GHA is presently broken again :()

jerboaa (approved):
It seems convoluted, TBH. The fix for OPENJDK-2408 would only need `rsync -rlt` (over `--archive`). I take it that would have the same problem? Anyway, it seems OK if it fixes the problem.

jmtd:
> It seems convoluted, TBH. The fix for OPENJDK-2408 would only need rsync -rlt (over --archive). I take it that would have the same problem?

Afraid so; `-t` is the issue (although other stuff that `--archive` tries to do would likely also fail, such as setting xattrs).

rsync has special behaviour when you suffix a `/` to a source path, which was a promising-looking fix:

> A  trailing slash on the source changes this behavior to avoid creating an additional directory level at the destination.  You can think of a trailing / on a source as meaning "copy the contents of this directory" as opposed to "copy the directory by name"

unfortunately it goes on to say 

> but in both cases the attributes  of  the  containing directory are transferred to the containing directory on the destination. 

The globstar is needed to match any hidden files in the source directory. I ran out of time to write an explicit test for that. The alternative `$srcdir/* $srcdir/.*` would match `.` and `..`.  The subshell (to avoid setting/altering globstar in the parent context) is just an abundance of caution, and might not strictly be necessary, but I didn't want to spend the time analysing the call chain to be certain.

> Anyway, it seems OK if it fixes the problem.

I think it does that, at least!

Thanks for the 👍🏼

PR #472, opened 2024-04-02
[OPENJDK-2735] Move grep and gawk installation to phase-1

jhuttana:
As suggested I have made an attempt to move `grep` and `gawk` installation to phase-1.
But with or without this change my builds are coming out with 
```
unable to retrieve container logs for cri-o:.........
```

 When I tried to check on web console as `adminstrator` I see that it is some ephemeral-storage issue.
```
Generated from kubelet on crc-vlf7c-master-0
The node was low on resource: ephemeral-storage. Threshold quantity: 4902142351, available: 4730208Ki. Container docker-build was using 4Ki, request is 0, has larger consumption of ephemeral-storage
```

I deleted all builds/pods/imagestreams but there is  no effect :)
Yesterday I was trying to create a deployment which was continuously failing and creating successive pods after every failure. I guessed that might have consumed all the memory and deleted all pods in that project. Now on local CRC I don't have any builds. Even then builds are failing with that `cri-o error`.

### DIFF
--- a/templates/jlink/jlink-builder/jlink-builder-template.yaml
+++ b/templates/jlink/jlink-builder/jlink-builder-template.yaml
@@ -29,6 +29,7 @@ objects:
       dockerfile: |
         FROM -
         USER 0
+        RUN microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y java-${JDK_VERSION}-openjdk-jmods
         RUN mkdir -p /mnt/jrootfs
         RUN microdnf install --installroot /mnt/jrootfs --releasever 9 --setopt install_weak_deps=0 --nodocs -y java-${JDK_VERSION}-openjdk-jmods \
         --config=/etc/dnf/dnf.conf \
--- a/templates/jlink/jlink-builder/jlink-builder-template.yaml
+++ b/templates/jlink/jlink-builder/jlink-builder-template.yaml
@@ -31,7 +31,7 @@ objects:
         USER 0
         RUN microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y java-${JDK_VERSION}-openjdk-jmods
         RUN mkdir -p /mnt/jrootfs
-        RUN microdnf install --installroot /mnt/jrootfs --releasever 9 --setopt install_weak_deps=0 --nodocs -y java-${JDK_VERSION}-openjdk-jmods \
+        RUN microdnf install --installroot /mnt/jrootfs --releasever 9 --setopt install_weak_deps=0 --nodocs -y \
         --config=/etc/dnf/dnf.conf \
         --noplugins \
         --setopt=cachedir=/var/cache \
--- a/templates/jlink/jlink-builder/jlink-builder-template.yaml
+++ b/templates/jlink/jlink-builder/jlink-builder-template.yaml
@@ -29,7 +29,15 @@ objects:
       dockerfile: |
         FROM -
         USER 0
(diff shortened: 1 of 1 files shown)


templates/jlink/jlink-builder/jlink-builder-template.yaml
```diff
@@ -29,7 +29,15 @@ objects:
       dockerfile: |
         FROM -
         USER 0
-        RUN microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y java-${JDK_VERSION}-openjdk-jmods
```
jmtd, line 32: This line shouldn't be changed: we need the jmods RPM to be installed in this container in the proper location.

templates/jlink/jlink-builder/jlink-builder-template.yaml
```diff
@@ -29,7 +29,15 @@ objects:
...
+        RUN mkdir -p /mnt/jrootfs
+        RUN microdnf install --installroot /mnt/jrootfs --releasever 9 --setopt install_weak_deps=0 --nodocs -y java-${JDK_VERSION}-openjdk-jmods \
```
jmtd: we don't need the jmods RPM content to be in the final image.

jmtd:
> When I tried to check on web console as adminstrator I see that it is some ephemeral-storage issue.

I hit this regularly with CRC. I would like to find a way to increase the amount of storage it has.

> I deleted all builds/pods/imagestreams but there is no effect :)

This hasn't worked for me either. When I hit this issue I have ot blow away the whole cluster and start a fresh (`crc delete`, `crc start`)

It looks like passing `-d` (Total size in GiB of the disk used by the instance (default 31)) to `crc start` will create a bigger disk.

templates/jlink/jlink-builder/jlink-builder-template.yaml
jhuttana: ok then I retain the original line and add the remaining stuff necessary for /mnt/jrootfs after the original line for jmods RPM

jmtd: Great -- that puts jmods back into the builder

templates/jlink/jlink-builder/jlink-builder-template.yaml
```diff
@@ -29,6 +29,7 @@ objects:
       dockerfile: |
         FROM -
         USER 0
+        RUN microdnf --setopt=install_weak_deps=0 --setopt=tsflags=nodocs install -y java-${JDK_VERSION}-openjdk-jmods
         RUN mkdir -p /mnt/jrootfs
         RUN microdnf install --installroot /mnt/jrootfs --releasever 9 --setopt install_weak_deps=0 --nodocs -y java-${JDK_VERSION}-openjdk-jmods \
```
jmtd: this line is still installing the jmods RPM into the jrootfs directory

jhuttana: Oops sorry :(  copy paste mistake.

jhuttana:
> > When I tried to check on web console as adminstrator I see that it is some ephemeral-storage issue.
> 
> I hit this regularly with CRC. I would like to find a way to increase the amount of storage it has.
> 
> > I deleted all builds/pods/imagestreams but there is no effect :)
> 
> This hasn't worked for me either. When I hit this issue I have ot blow away the whole cluster and start a fresh (`crc delete`, `crc start`)
> 

In my case even stop and start didnt work :D 
I just reconfigured crc and then it worked.

> It looks like passing `-d` (Total size in GiB of the disk used by the instance (default 31)) to `crc start` will create a bigger disk.

That will help.

jmtd (approved):

PR #473, opened 2024-04-03
Disable a flaky test on CI

jmtd:
This test is unreliable, it provokes some kind of race in behave-test-steps and sometimes will block indefinitely.

### DIFF
--- a/modules/jvm/tests/features/runtime.feature
+++ b/modules/jvm/tests/features/runtime.feature
@@ -52,6 +52,8 @@ Feature: Openshift OpenJDK Runtime tests
     | JAVA_APP_DIR | .      |
   Then available container log should contain INFO running in /home/default
 
+  # too flaky for CI
+  @ignore
   @ubi9
   Scenario: Check non-existent path JAVA_APP_DIR (OPENJDK-2034)
     Given container is started with env


jmtd:
They are now blocking on a different test. Example from JDK11
```
Feature: Miscellaneous OpenJDK-related unit tests # features/jboss.container.openjdk/openjdk.feature:1

  @ubi9/openjdk-11 @ubi9/openjdk-11-runtime
  Scenario: Check that only OpenJDK 11 is installed            # features/jboss.container.openjdk/openjdk.feature:5
    When container is started with args                        # steps/container_steps.py:86
      | arg     | value   |
      | command | rpm -qa |
2024-04-03 15:12:15,643 container.py:238        DEBUG Creating container from image 'ubi9/openjdk-11-runtime:1.18'...
2024-04-03 15:12:15,643 container.py:254        DEBUG Creating docker container with arguments and image:  ubi9/openjdk-11-runtime:1.18
2024-04-03 15:12:15,658 container.py:95         DEBUG Starting container 'c0e7056c158b85a1bebefec30cbd5006b82c22280f52d8142fbf65db4448f37c'...
2024-04-03 15:12:15,832 container_steps.py:317        DEBUG Running command ps -C java in container
2024-04-03 15:12:15,832 container.py:152        DEBUG container.execute(ps -C java,False)
Error: The operation was canceled.
```


The underlying issue is that the bits of `behave-test-steps` we are using are unreliable/jittery. Possibly worse after merging https://github.com/cekit/behave-test-steps/pull/50 , rather than better; I'm not sure.

The solution has to be making `behave-test-steps` more robust; trying to address this with workarounds in the image repositories is doomed to fail.

PR #474, opened 2024-04-04
OPENJDK-2734: Stage 2 initial implementation

Josh-Matsuoka:
Initial implementation of build stage 2, addresses OPENJDK-2734

Leaving as draft for the moment, I'll do a second pass over it tomorrow. Wanted to get this out ahead of the meeting.

### DIFF
--- a/templates/jlink/jlink-builder/jlink-builder-template.yaml
+++ b/templates/jlink/jlink-builder/jlink-builder-template.yaml
@@ -10,6 +10,15 @@ parameters:
   name: JDK_VERSION
   value: "11"
   required: true
+- description: Application to run the jlink workflow on
+  name: APPLICATION
+  required: true
+- description: Git Ref to run the jlink workflow on
+  name: REF
+  required: true
+- description: Context Dir to use for the jlink workflow
+  name: CONTEXT_DIR
+  required: true
 objects:
 - apiVersion: image.openshift.io/v1
   kind: ImageStream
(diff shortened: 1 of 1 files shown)


templates/jlink/jlink-builder/jlink-builder-template.yaml
```diff
@@ -49,3 +58,31 @@ objects:
       to:
         kind: ImageStreamTag
         name: ubi9-openjdk-${JDK_VERSION}-jlink:latest
+- apiVersion: build.openshift.io/v1
+  kind: BuildConfig
```
jerboaa, line 69: `BuildConfig` is OpenShift specific. It would be best to design this based on Builds for Red Hat OpenShift which is k8s specific which covers a broader range of deployments and would be more forward-looking:
https://access.redhat.com/documentation/en-us/builds_for_red_hat_openshift/1.0

jmtd: I think we need to move to `Build` instead of `buildConfig`, yes (also in #471). I won't block merging this to `jlink-dev` for that reason, though, it goes on the TODO list.

templates/jlink/jlink-builder/jlink-builder-template.yaml
```diff
@@ -49,3 +58,31 @@ objects:
...
+        pullPolicy: Always
+      env:
+        - name: S2I_ENABLE_JLINK
+          value: true
+        - name: LOGGING_SCRIPT_DEBUG
+          value: true
+  output:
+    to:
+      kind: ImageStreamTag
+      name: ubi9-openjdk-${JDK_VERSION}-jlink:latest
```
jmtd: this is the input `imageStream`. This phase needs to output to a different `imageStream`, which will also need adding to the template. The name could include `intermediate` in it. for a first cut that would be enough.  Later it should probably also be unique to the app in some way. We might need to add a template parameter that the user can provide that we incorporate into the name. But I'd put that into another work stream and not block this PR on it.

templates/jlink/jlink-builder/jlink-builder-template.yaml
```diff
@@ -49,3 +58,31 @@ objects:
...
+- apiVersion: build.openshift.io/v1
+  kind: BuildConfig
+  metadata:
+    name: jlink-s2i-jdk-${JDK_VERSION}
+    labels:
+      app: jlink-s2i-jdk-${JDK_VERSION}
+  spec:
+    source:
+      type: Git
+        git:
```
jmtd: these four lines (70-73) need their indent decrementing (git and contextDir should be keys under `source` not under `type`)

jmtd:
@Josh-Matsuoka , thanks for raising this PR. This is great! I've just done a thorough test of this  and an attempt at combining it with the other templates to run the whole thing through end-to-end. I've just pushed two commits to your branch/this PR to resolve some immediate syntax problems. I'm just putting together the next steps we need to do and I can cover that when we meet later.

PR #397, opened 2023-09-14
[OPENJDK-2063] Tweak run scripts to support docker multistage build for jlinked JRE container application

jhuttana:
https://issues.redhat.com/browse/OPENJDK-2063

This PR is to address GH issues 
https://github.com/jboss-container-images/openjdk/issues/394
https://github.com/jboss-container-images/openjdk/issues/395

which are some sub-tasks for above mentioned JIRA tracker.

I haven't removed the definition of JBOSS_CONTAINER_UTIL_LOGGING_MODULE from modules/util/logging/module.yaml because it is referenced in modules/maven/* and modules/s2i/* in multiple files. So I feel its not a good idea to define it in all those files.

PR #412, opened 2023-11-07
[OPENJDK-2200] Test coverage for user-provided run-env.sh

jhuttana:
JIRA issue: https://issues.redhat.com/browse/OPENJDK-2200
GH issue: https://github.com/jboss-container-images/openjdk/issues/399
The referenced application source is included in the PR : https://github.com/jboss-container-images/openjdk-test-applications/pull/3

PR #478, opened 2024-04-08
[OPENJDK-2915] refactor run-env.sh into run-java.sh, clean up run script

jmtd:
<https://issues.redhat.com/browse/OPENJDK-2915>

This is some clean-up work that was needed prior to implementing the
jlink integration scripts.

The file `run-env.sh` is shipped alongside `run-java.sh` in the containers.
All it did was to set a default value for JAVA_APP_DIR, if it was not
already set.

There is a related feature of the run script: if a run-env.sh script
is present within the application dir (JAVA_APP_DIR), then it is
evaluated. Therefore an application source can extend the run behaviour
with extra shell script.

This clean-up removes the `run-env.sh` that is shipped alongside `run-java.sh`,
moving what it directly into `run-java.sh` instead. Via separate PRs, we have
added unit test coverage for the behaviours listed above:

1. [Ensure that run-env.sh placed in the JAVA_APP_DIR is sourced in the run script before launching java](https://github.com/jboss-container-images/openjdk/blob/ubi9/modules/maven/s2i/tests/features/java_s2i.feature#L178)
2. [test coverage for setting values of JAVA_APP_DIR](https://github.com/jboss-container-images/openjdk/blob/ubi9/modules/jvm/tests/features/runtime.feature)

----

This is a port of https://github.com/jboss-container-images/openjdk/pull/397 for the ubi9 branch. I hope to merge this (and another piece) prior to looking at `jlink-dev`, to reduce the size of the jlink delta and make review easier.

This is @jhuttana 's work.

### DIFF
--- a/modules/run/artifacts/opt/jboss/container/java/run/run-env.sh
+++ b/modules/run/artifacts/opt/jboss/container/java/run/run-env.sh
@@ -1,6 +1,0 @@
-#!/bin/sh
-
-# Default the application dir to the S2I deployment dir
-if [ -z "$JAVA_APP_DIR" ]
-then JAVA_APP_DIR=/deployments
-fi
--- a/modules/run/artifacts/opt/jboss/container/java/run/run-java.sh
+++ b/modules/run/artifacts/opt/jboss/container/java/run/run-java.sh
@@ -4,8 +4,8 @@
 set -eo pipefail
 
 #These are defined explicitly here to avoid defining them in templates/jlink/Dockerfile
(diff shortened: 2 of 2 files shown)


jerboaa (approved):
Seems fine.

jhuttana:
We already have test coverage for this on ubi9 also 
https://github.com/jboss-container-images/openjdk/blob/ubi9/modules/maven/s2i/tests/features/java_s2i.feature#L178
PR : https://github.com/jboss-container-images/openjdk/pull/412

PR #479, opened 2024-04-10
[OPENJDK-2837] UBI9: Rename the test suitably

jhuttana:
I have renamed the test to `imagebasic.feature` as it tests some basic stuffs of the created image.


jmtd:
I left some review notes on the UBI8 PR which apply here: https://github.com/jboss-container-images/openjdk/pull/480#issuecomment-2051361053

Mainly, I'm not sure what the reason is to rename it.

jhuttana:
> I left some review notes on the UBI8 PR which apply here: [#480 (comment)](https://github.com/jboss-container-images/openjdk/pull/480#issuecomment-2051361053)
> 
> Mainly, I'm not sure what the reason is to rename it.


Actually you suggested to rename this file with the reason: "The label checks at the beginning test stuff in the image descriptors rather than modules; they should stay where they are for now (although they aren't "openshift" tests so we should probably rename the file to another name within tests/features)"

jmtd (approved):

PR #484, opened 2024-04-15
[OPENJDK-2824] assemble: binary: Don't set times on directories (UBI8)

jmtd:
If the S2I build runs as a user other than 185 (such as s2i's default
of 1001), attempting to set the timestamp of /deployments will fail,
causing the build to fail.

https://issues.redhat.com/browse/OPENJDK-2824

See also

Backport of #470


jmtd:
Running tests locally...

jmtd:
New test passes. Running the whole suite now

jerboaa (approved):

jmtd:
whole suite passes

PR #489, opened 2024-04-29
[OPENJDK-2899]Add hooks for stage-2 build

jhuttana:
JIRA: https://issues.redhat.com/browse/OPENJDK-2899
Could you please take a look?

I haven't created the webhook yet.

### DIFF
--- a/templates/jlink/jlinked-app.yaml
+++ b/templates/jlink/jlinked-app.yaml
@@ -21,6 +21,12 @@ parameters:
 - description: Context Dir to use for the jlink workflow
   name: CONTEXT_DIR
   required: true
+- description: A secret string used to configure the GitHub webhook
+  name: GITHUB_WEBHOOK_SECRET
+  generate: expression
+  from: "[a-zA-Z0-9]{16}"
+  required: true
+message: "... The GitHub webhook secret is ${GITHUB_WEBHOOK_SECRET} ..."
 ##############################################################################
 # Following are the objects(imagestream and buildconfigs) for all the 3-stages
 objects:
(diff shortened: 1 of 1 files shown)


templates/jlink/jlinked-app.yaml
```diff
@@ -105,6 +105,17 @@ objects:
...
+    triggers:
+      - type: ConfigChange
+      - type: ImageChange
+        imageChange:
+          from:
+            kind: ImageStreamTag
+            name: ubi9-openjdk-${JDK_VERSION}-jlink:latest  # Output of stage-1 which serves as input to stage-2
+      - type: GitHub
+        github:
+          secret: Gvw2O0tXuaAUxleJ
```
jmtd: pLease have a quick investigation into whether we can replace this secret with a template parameter, and if it’s possible for that to default to a random string.

jmtd: See this example https://github.com/jboss-container-images/openjdk/blob/b2201a5f43ff161db8e29bea4f5807de278400cf/templates/openjdk-web-basic-s2i.json#L66

jhuttana:
I have made the suggested changes to generate random value for secret.
Also, verified that the value is getting substituted appropriately in the buildConfig after template is instantiated.

```
$ oc process --parameters=true jlink-app-template
NAME                    DESCRIPTION                                                GENERATOR           VALUE
JDK_VERSION             JDK version to produce a jmods image and imagestream for                       11
APP_URI                 Application to run the jlink workflow on                                       
REF                     Git Ref to run the jlink workflow on                                           
CONTEXT_DIR             Context Dir to use for the jlink workflow                                      
GITHUB_WEBHOOK_SECRET   A secret string used to configure the GitHub webhook       expression          [a-zA-Z0-9]{16}
```
```
$  oc get bc
NAME                     TYPE     FROM         LATEST
jlink-builder-jdk-17     Docker   Dockerfile   0
jlink-s2i-jdk-17         Source   Git@master   0
multistage-buildconfig   Docker   Dockerfile   1
```

```
$ oc get -o yaml bc/jlink-s2i-jdk-17
.
.
.
triggers:
.
.
.
  - github:
      secret: REuvkipBJ3aWOFEE
    type: GitHub
status:
  lastVersion: 0
```

jmtd (approved):
Fantastic, thanks!

PR #492, opened 2024-05-10
[Openjdk-2969] runtime images and JAVA_ARGS, JAVA_OPTIONS handling (ubi8)

jmtd:
Some of the run logic is implemented in bin/s2i which means it is not applied in the runtime images.

https://issues.redhat.com/browse/OPENJDK-2969


jmtd:
I don't think we'll ever land this in ubi8. See #589 for  the ubi9 version.

PR #497, opened 2024-05-23
[OPENJDK-3009] syntax fix for run-java.sh

jmtd:
The backporting of mask_passwords to UBI8 had a syntax error, we borrowed the closing parenthesis from mask_passwords(), breaking the run-java.sh script.

https://issues.redhat.com/browse/OPENJDK-3009

I think I introduced this when undoing my attempt to remove `configure_passwd` first time around :(


sefroberg (approved):
Looks good. I run bash -n on the file I change and it will catch this.

jerboaa (approved):

PR #498, opened 2024-05-30
[Openjdk 3029] no singleton jdk - maven module edition

jmtd:
Removing singleton-jdk from the JDK21 builder image is not sufficient, we also need to install `maven-openjdk21` instead of `maven-openjdk17`. Thanks @sefroberg for catching this.

Test coverage: the JDK21 tests in `modules/jdk/tests/features/openjdk.feature` catch this (pass for this PR, fail for branch ubi8)

https://issues.redhat.com/browse/OPENJDK-3029

### DIFF
--- a/modules/maven/21/configure.sh
+++ b/modules/maven/21/configure.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+set -e
+
+# This file is shipped by a Maven package and sets JAVA_HOME to
+# an OpenJDK-specific path. This causes problems for OpenJ9 containers
+# as the path is not correct for them.  We don't need this in any of
+# the containers because we set JAVA_HOME in the container metadata.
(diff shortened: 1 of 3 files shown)


sefroberg (approved):
Looks good.

sefroberg:
Two Questions.
1. I noticed a Maven update to 3.8.21. Should we do this across the board for all containers?
2. Are you going to make this change for the UBI 9 images as well?

jmtd:
> I noticed a Maven update to 3.8.21. Should we do this across the board for all containers?

I guess that came in with RHEL 8.10. I think/hope it should be consistent across the builds as they happened at the same time. But we should check.
 
Previous/current version in both the UBI8 and UBI9 images is 3.8.5

PR #500, opened 2024-06-10
[OPENJDK-2992] Spec for deployment and route object

jhuttana:
JIRA: https://issues.redhat.com/browse/OPENJDK-2992

NOTE: My stage-2 is failing with CRI-O failed msg which is known one with our crc setup. I will have to restart or do a fresh setup to get it working fine :) Meanwhile I thought I can raise a PR so that you can take look and commented if its syntactically fine.

### DIFF
--- a/templates/jlink/jlinked-app.yaml
+++ b/templates/jlink/jlinked-app.yaml
@@ -213,3 +213,43 @@ objects:
           from:
             kind: ImageStreamTag
             name: ubimicro:latest  # ImageStreamTag for registry.access.redhat.com/ubi9/ubi-micro
+##############################################################################
+# Deployment and Route object specs
+- apiVersion: apps.openshift.io/v1
+  kind: Deployment
+  metadata:
+    name: jlinked-app
+  spec:
+    replicas: 1
+    selector:
+      matchLabels:
+        app: jlinked-app
+    template:
+      metadata:
+        labels:
+          app: jlinked-app
+      spec:
+        containers:
+          - name: jlinked-app-container
+            image: 'openshift/jlinked-app:latest'
+            ports:
+              - containerPort: 8080
+- apiVersion: v1
+  kind: Service
+  metadata:
+    name: jlinked-app-service
+  spec:
+    selector:
+      app: jlinked-app
+    ports:
+      - protocol: TCP
+        port: 80
+        targetPort: 8080
+- apiVersion: route.openshift.io/v1
+  kind: Route
+  metadata:
+    name: jlinked-app-route
+  spec:
+    to:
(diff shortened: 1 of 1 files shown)


jmtd:
I've just tried this out.

the deployment object starts to log errors as soon as the template is instantiated, and before the builds have completed.

Once the builds have completed, the route is not yet working (Application is not available)

I'm creating a deployment manually now to compare the objects in the OCP instance

jmtd:
app
one difference, the manual deployment has some kind of "Application" enclosing it

jmtd:
I think the route object needs `.spec.port.targetPort: 8080-tcp` added (possibly also via a template variable)

templates/jlink/jlinked-app.yaml
```diff
@@ -213,3 +213,44 @@ objects:
...
+      matchLabels:
+        app: jlinked-app
+    template:
+      metadata:
+        labels:
+          app: jlinked-app
+      spec:
+        containers:
+          - name: jlinked-app-container
+            image: lightweight-image:latest
```
jmtd, line 234: I think this might be an issue. Comparing this to a manually-created Deployment, the working one has, instead,

    - image: image-registry.openshift-image-registry.svc:5000/jlink2/lightweight-image@sha256:19538848d2f06ee00d7462f8290982191fe9acb63c6ecea026d8df484e840ecc

jmtd: ...but that's also specific to my OpenShift cluster. I'm not sure what to put in a template. re-reading https://docs.openshift.com/container-platform/4.15/openshift_images/using-templates.html, the example they have of a template with a Deployment in it does not name the image at all.

jhuttana: Actually I referred this template : https://docs.openshift.com/container-platform/4.15/applications/deployments/what-deployments-are.html#deployments-kube-deployments_what-deployments-are 
DeploymentConfig is marked as deprecated and they recommend to use Deployment object.

jmtd: I hadn't noticed the discrepancy between Deployment and DeploymentConfig. Hmm. I see (from that example) that DeploymentConfig supports specifying an ImageStreamTag. It  doesn't look like Deployment does (which makes sense, since Deployment comes from Kubernetes but ImageStreams are an OpenShift addition). So I'm not really sure how we would specify the image to use for the Deployment in a safe/generic way.

Even though DCs are deprecated, I think I'd be happy to merge a PR which used them if it worked today!

jmtd:
Work on this continues in #514.

PR #502, opened 2024-07-04
[OPENJDK-2903] [jlink] Ensure created objects are uniquely named

Josh-Matsuoka:
Hi,

This PR fixes OPENJDK-2903, adding a mandatory template parameter (APPNAME) that ensures all created resources are uniquely named.

https://issues.redhat.com/browse/OPENJDK-2903

### DIFF
--- a/templates/jlink/jlinked-app.yaml
+++ b/templates/jlink/jlinked-app.yaml
@@ -199,7 +199,7 @@ objects:
       dockerStrategy:
         from:
           kind: ImageStreamTag
-          name: ubimicro:latest
+          name: ${APPNAME}-ubimicro:latest
     output:
       to:
         kind: ImageStreamTag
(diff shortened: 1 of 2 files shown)


jmtd (changes requested):
Looks mostly good, thanks!

There are a couple of references to the ImageStream `ubimicro:latest` which need updating to be prefixed with the template variable.

jmtd:
@Josh-Matsuoka please add reference to `APPNAME` in `templates/jlink/README.md` too!

templates/jlink/jlinked-app.yaml
```diff
@@ -44,9 +47,9 @@ objects:
 - apiVersion: build.openshift.io/v1
   kind: BuildConfig
   metadata: 
-    name: jlink-builder-jdk-${JDK_VERSION}
+    name: ${APPNAME}-jlink-builder-jdk-${JDK_VERSION}
     labels:
-      app: jlink-builder-jdk-${JDK_VERSION}
+      app: ${APPNAME}-jlink-builder-jdk-${JDK_VERSION}
```
jmtd, line 52: I wonder if we could just use `label.app: ${APPNAME}` throughout here… once my CRC instance rebuilds I'll see what it looks like

jmtd: let's not block on that

PR #516, opened 2024-11-07
[OPENJDK-3419] Initial changes for JDK11-ELS images

vieiro:
Update github workflow and README.adoc

Signed-off-by: Antonio Vieiro <avieirov@redhat.com>

### DIFF
--- a/.github/workflows/openjdk-11-rhel9.yaml
+++ b/.github/workflows/openjdk-11-rhel9.yaml
@@ -2,9 +2,9 @@ name: UBI9 OpenJDK 11 S2I Image CI
 on: [push, pull_request]
 env:
   LANG: en_US.UTF-8
-  IMAGE: ubi9-openjdk-11
+  IMAGE: openjdk-11-rhel-9
 jobs:
   call-openjdkci:
     uses: ./.github/workflows/image-workflow-template.yml
     with:
-      image: ubi9-openjdk-11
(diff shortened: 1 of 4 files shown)


jerboaa:
GHA fails with

```
ERROR Descriptor ('openjdk-11-rhel-9.yaml') could not be found on the path, please check your arguments!
```

The descriptor in this patch is `openjdk-11-rhel9.yaml`

vieiro:
Hi @jmtd @sefroberg,
Ready for review now. I've commented out the failing `behave` tests (we may improve them later on).
Apologies for the force pushes (this is something I'm used to in other projects where squashes are not welcome).

sefroberg (approved):
Assuming we figure out the behave tests next week, I think this looks correct.

.github/workflows/openjdk-11-rhel9.yaml
```diff
@@ -2,9 +2,9 @@ name: UBI9 OpenJDK 11 S2I Image CI
 on: [push, pull_request]
 env:
   LANG: en_US.UTF-8
-  IMAGE: ubi9-openjdk-11
+  IMAGE: openjdk-11-rhel-9
```
jmtd, line 5: It would seem this PR proves that this env var definition here is not in use.

vieiro: Very true!

PR #518, opened 2024-11-12
[OPENJDK-3426] Adjust Behave test tagging for new container names

vieiro:
- Drops 17 & 21.
- Adjusts the test tagging to `@openjdk-els/openjdk-11-rhel9`
- Re-enables the GHA tests.

See https://issues.redhat.com/browse/OPENJDK-3426


jmtd:
Hi @vieiro , the changes look correct, and the gHA failure is a replication of what we currently see on the main branch as well, and not  a problem with this PR.

My only comment would be, whether we should consider if removing the unrelated tests and untagging tests for the other versions will cause potential merge conflicts with cherry picking commits in the future. The spare tests and tags are otherwise benign. If we add a test you want or fix a bug or similar in an existing test, it may require manual merging due to the nearby tag changes.

In fact we could probably apply a patch to the main branch which rename the test tags (and merge it after we EOL the main JDK11 images), to reduce the difference between the main and els branches in the future. WDYT?

vieiro:
Hi @jmtd sounds perfectly reasonable. Will redo and force-push.

vieiro:
Hi @jmtd ,
Does this look any better?

vieiro:
I'm not sure about dropping 17 & 21 `.yml` files, though. 
Shall we keep them (even though we still drop the GHA files for 17 & 21)?

jmtd:
I think this looks better.

Dropping the GHA files for 17 & 21 is fine, because if we make any changes to the GHA setup upstream, it will be to `.github/workflows/image-workflow-template.yml`, rather than to those.

As for the image descriptors, best leave them. I did an experiment: remove them, then cherry-pick 374c13bda1a76baead6d8077f75aeeb0f46a8e08 (from #519 ), the result was merge conflicts for those files. (unfortunately merge conflicts for the JDK11 files too, but that can't be avoided). By leaving the files in-tree, most of those conflicts are avoided.

vieiro:
@jmtd So we just drop the GHA files for 17 & 21 now, and keep the image descriptors.

I tried a cherry-pick of https://github.com/rh-openjdk/redhat-openjdk-containers/commit/374c13bda1a76baead6d8077f75aeeb0f46a8e08 and yes, we only have the unavoidable conflicts for 11 image descriptors.

jmtd (approved):

PR #519, opened 2024-11-12
[OPENJDK-3430] alter FROM line for redhat builds

jmtd:
In the redhat/* overrides files, override "from:" to use the name assigned to RHEL images built via OSBS from 9.5 onwards.

https://issues.redhat.com/browse/OPENJDK-3430


jmtd:
We have a successful internal build from this change now. Confirming that it has picked up the correct parent image.

jmtd:
I've done two internal builds and confirmed they have picked up the correct base images.

sefroberg (approved):
LGTM

PR #520, opened 2024-11-18
[OPENJDK-3414] Update refs to openjdk-test-applications

jmtd:
The repository has moved: update references.

https://issues.redhat.com/browse/OPENJDK-3414


sefroberg (approved):
Looks good to me. I do wonder if there is a way with these tests to pass an environment variable into it so that we only have to make this change once.

jmtd:
Thanks for approving!

> I do wonder if there is a way with these tests to pass an environment variable into it so that we only have to make this change once.

At the moment there isn't but we could add something like that by overhauling [behave-test-steps](https://github.com/cekit/behave-test-steps).

PR #545, opened 2025-02-18
[OPENJDK-3654] Remove superseded assemble script

jmtd:
This script is overwritten by another later in the build stage for all our images. The "winning" script is
"modules/s2i/bash/artifacts/usr/libexec/s2i/assemble".
https://issues.redhat.com/browse/OPENJDK-3654


sefroberg (approved):
LGTM

PR #548, opened 2025-02-18
[OPENJDK-3652] don't try to set alternatives for JDK

jmtd:
The exact invocation has changed for RHEL10. This no longer works:

    alternatives --set java java-21-openjdk.${_arch}

This works, but doesn't in earlier RHELs:

    alternatives --set java /usr/lib/jvm/java-21-openjdk/bin/java

Some background in RHEL-68848

However, there is only one JDK in RHEL10 at the moment anyway, so simply avoid adjusting alternatives.

https://issues.redhat.com/browse/OPENJDK-3652

### DIFF
--- a/modules/jdk/21/configure.sh
+++ b/modules/jdk/21/configure.sh
@@ -13,13 +13,6 @@ pushd ${ARTIFACTS_DIR}
 cp -pr * /
 popd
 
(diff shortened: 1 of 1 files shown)


sefroberg (approved):
Looks good to me.

PR #550, opened 2025-02-18
[OPENJDK-3659] remove python from ubi10 images

jmtd:
Hopefully temporary, see   https://issues.redhat.com/browse/RHEL-79954
https://issues.redhat.com/browse/OPENJDK-3659

### DIFF
--- a/modules/util/rm-python/execute.sh
+++ b/modules/util/rm-python/execute.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+set -euo pipefail
+
+rpm -e $(rpm -qa python*)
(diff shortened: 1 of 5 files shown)


sefroberg:
@jmtd Let's put a hold on this. I don't think removing python, if it should be installed is a good idea. I have reached out to the RHEL folks on our internal Slack for more guidance.

jmtd:
> I don't think removing python, if it should be installed is a good idea.

It's not clear why it's installed. None of the other installed RPMs depend upon it.

>  I have reached out to the RHEL folks on our internal Slack for more guidance

Can you link me to that chat please? I'm keen to see what they say.

jmtd:
resolved upstream

PR #552, opened 2025-02-18
[OPENJDK-3634] update image name in jlink template

Josh-Matsuoka:
Addresses https://issues.redhat.com/browse/OPENJDK-3634

Brings over the work done in https://github.com/rh-openjdk/redhat-openjdk-containers/pull/514/commits/884c6212c36324118dc8198572917fe2175cec11 , Updates the image name in jlink template to reflect it's tech-preview status and the new name of the jlink container. Also bumps up the recommended/example JDK version to 21.

### DIFF
--- a/templates/jlink/README.md
+++ b/templates/jlink/README.md
@@ -16,11 +16,11 @@ prepare UBI9 OpenJDK ImageStreams with `jlink-dev` support.
    repository](https://github.com/jboss-container-images/openjdk),
    branch `jlink-dev`. e.g.
 
-        cekit --descriptor ubi9-openjdk-17.yaml build docker
+        cekit --descriptor ubi9-openjdk-21.yaml build docker
 
 2. Within your OpenShift project,
 
-        oc create imagestream ubi9-openjdk-17
+        oc create imagestream ubi9-openjdk-21-jlink-tech-preview
 
 3. You may need to configure your container engine to not TLS-verify the OpenShift
    registry. For Docker, add the following to `/etc/docker/daemon.json` and restart
@@ -37,8 +37,8 @@ prepare UBI9 OpenJDK ImageStreams with `jlink-dev` support.
 5. tag and push the dev image into it. The OpenShift console gives you the
    exact URI for your instance
 
-        docker tag ubi9/openjdk-17:1.18 default-route-openshift-image-registry.apps-crc.testing/jlink1/ubi9-openjdk-17:1.18
-        docker push default-route-openshift-image-registry.apps-crc.testing/jlink1/ubi9-openjdk-17:1.18
+        docker tag openjdk-tech-preview/openjdk-21-jlink-rhel9:1.18 default-route-openshift-image-registry.apps-crc.testing/jlink1/ubi9-openjdk-21-jlink-tech-preview:1.18
+        docker push default-route-openshift-image-registry.apps-crc.testing/jlink1/ubi9-openjdk-21:1.18
 
 ## Stage 1: Load the template into OpenShift and instantiate it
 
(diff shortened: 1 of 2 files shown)


templates/jlink/README.md
```diff
@@ -37,8 +37,8 @@ prepare UBI9 OpenJDK ImageStreams with `jlink-dev` support.
 5. tag and push the dev image into it. The OpenShift console gives you the
    exact URI for your instance
 
-        docker tag ubi9/openjdk-17:1.18 default-route-openshift-image-registry.apps-crc.testing/jlink1/ubi9-openjdk-17:1.18
-        docker push default-route-openshift-image-registry.apps-crc.testing/jlink1/ubi9-openjdk-17:1.18
+        docker tag openjdk-tech-preview/openjdk-21-jlink-rhel9:latest default-route-openshift-image-registry.apps-crc.testing/jlink1/ubi9-openjdk-21-jlink-tech-preview:latest
+        docker push default-route-openshift-image-registry.apps-crc.testing/jlink1/ubi9-openjdk-21:latest
```
jmtd: the last argument to docker push needs to match the last argument to the previous docker tag -- pls update to include the tech-preview bit

templates/jlink/jlinked-app.yaml
```diff
@@ -78,18 +78,18 @@ objects:
       dockerStrategy:
         from: 
           kind: ImageStreamTag
-          name: ubi9-openjdk-${JDK_VERSION}:${BUILDER_IMAGE_TAG}  # Refer README.md to create this ImageStream
+          name: openjdk-${JDK_VERSION}-jlink-tech-preview:${BUILDER_IMAGE_TAG}  # Refer README.md to create this ImageStream
```
jmtd, line 81: this doesn't match the name of the imagestream in README (which is `ubi9-openjdk-21-jlink-tech-preview` or `ubi9-openjdk-${JDK_VERSION}-jlink-tech-preview`)

jmtd (changes requested):
see inline messages

templates/jlink/jlinked-app.yaml
Josh-Matsuoka: Right, in this case the README should be updated.

sefroberg:
We should not be over writing the existing ubi9-jdk21 image but adding a new file here. I don't see how what we are doing is going to save us any time or effort in the short, medium or long-term. In the long term, it will make attempting to merge this branch back into the UBI9 branch impossible.

jmtd:
Hi @sefroberg 

can you expand on what the problems are?

> In the long term, it will make attempting to merge this branch back into the UBI9 branch impossible.

In the future if we are looking at merging `jlink-dev` down to `ubi9`, that is presumably because we are moving the feature from Tech Preview to GA. In which case, we'd want the changes in the main UBI9/21 image, rather than in a separate TP image, which would be going away at that point.  We would need to revert the image name changes at that point, and make various other changes. But in either case the changes needed to the image YAML are tiny.

jmtd:
Also to be clear the image changes are already merged, happy to have that discussion but this PR doesn't touch the image YAMLs.

jmtd:
@Josh-Matsuoka I'm reviewing this now but please take a look at resolving merge conflicts: something has changed in `jlink-dev` since you forked for this. Thanks!

templates/jlink/jlinked-app.yaml
```diff
@@ -41,7 +41,7 @@ objects:
 - apiVersion: image.openshift.io/v1
   kind: ImageStream
   metadata:
-    name: ${APPNAME}-ubi9-openjdk-${JDK_VERSION}-jlink
+    name: ${APPNAME}-openjdk-${JDK_VERSION}-jlink-tech-preview
```
jmtd, line 44: this didn't need to change, but, so long as it's consistent in the template, that doesn't matter. (looks consistent to me)

jmtd (approved):
LGTM (subject to resolving merge conflicts)

PR #554, opened 2025-02-19
OPENJDK-3650 centos streams

jmtd:
This allows us to work close to RHEL cutting edge (whereas RHEL 10 beta is frozen in time) and to work in public, with e.g. GHA able to build and (behave) test images.

https://issues.redhat.com/browse/OPENJDK-3650

### DIFF
--- a/.github/workflows/image-workflow-template.yml
+++ b/.github/workflows/image-workflow-template.yml
@@ -17,7 +17,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Verify latest UBI image is present
-        run:  docker pull registry.access.redhat.com/ubi9/ubi-minimal:latest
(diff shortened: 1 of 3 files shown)


jmtd:
@sefroberg would you have time to take another look at this please?

sefroberg (approved):
We just need to keep inmind there is a delta from the centos stream images and the rhel 10 one. I am ok with this change. Looks good to me.

jmtd:
>  We just need to keep inmind there is a delta from the centos stream images and the rhel 10 one. I am ok with this change. Looks good to me.

Yes you are right: for example I don't think the Python problem exists in RHEL10 (yet). Towards the end of the dev cycle we need to switch (back) and resolve any discrepancies. I'll try to keep an eye on the delta in the meantime.

PR #558, opened 2025-02-24
[OPENJDK-3665] Remove maven_s2i_source_maven_overrides (not used)

jmtd:
This was a scheme to permit a downstream container to override the behaviour of maven_s2i_init, but in practise it's unused and just complicates our scripts.
https://issues.redhat.com/browse/OPENJDK-3665


sefroberg (approved):
Looks good to me.

PR #559, opened 2025-02-24
[OPENJDK-3676] convert installing some things over to artifacts:

jmtd:
This would resolve https://issues.redhat.com/browse/OPENJDK-2814 and https://issues.redhat.com/browse/OPENJDK-3674

----

The pattern used in our containers for copying files in, up to now, was to do so entirely inside a script `configure.sh`, with the files to be copied in an adjacent directory `artifacts`, normally with a local directory heirarchy mirroring the destination in the image, e.g. `modules/jdk/21/artifacts/opt/jboss/container/openjdk/jdk/jvm-options`. The script usually changed the ownership and/or permissions of the files after copying them.

Cekit offers a different method, [artifacts](https://docs.cekit.io/en/latest/descriptor/module.html#artifacts), which for some reason we overlooked using for this. Using this is normally shorter, by default sets the owner and group to root, and is more declarative, so makes the images easier to reason about. For the above example artifact, the diff to the configuration to copy it in would be

```yaml
artifacts:
  - path: artifacts/opt/jboss/container/openjdk/jdk/jvm-options
    dest: /opt/jboss/container/openjdk/jdk
```

I've gone a step further in this PR and removed the deep directory heirarchies encoding the destination path, since that's now present in the YAML. Personally I find the deep directory heirarchies hard to work with.

Most of the `configure.sh` scripts have been removed, as all they did was the copying.

The copied files are all now owner by root:root, which I prefer, it was always weird to me that the scripts could theoretically be modified by the running web application.

Cekit `artifacts` handling does not change file permissions, so I've set the necessary permissions on the scripts in git itself.


sefroberg (approved):
Ok, after reading the artifacts doc I think I get it. I wonder if structuring the github repo to mimic the container image's structure might make this whole operation a lot more simple. Great progress though. For what you are attempting to do here, great improvements. Looks good to me.

PR #561, opened 2025-02-25
[OPENJDK-2815] stop creating '*' symlink

jmtd:
This was an attempt at creating a fixed path irrespective of Maven version but it has never worked, and instead it created a symlink with the basename '*'.

https://issues.redhat.com/browse/OPENJDK-2815


sefroberg (approved):
Looks good to me.

PR #562, opened 2025-02-25
[OPENJDK-3674] update references to maven 3.8 to 3.9

jmtd:
The Maven provided in RHEL10 is Maven 3.9.

https://issues.redhat.com/browse/OPENJDK-3674

### DIFF
--- a/ubi10-openjdk-21.yaml
+++ b/ubi10-openjdk-21.yaml
@@ -51,7 +51,7 @@ modules:
   - name: jboss.container.openjdk.jdk
     version: "21"
   - name: jboss.container.maven
-    version: "3.8.21"
+    version: "3.9.21"
   - name: jboss.container.util.tzdata
   - name: jboss.container.java.s2i.bash
   - name: jboss.container.util.tzdata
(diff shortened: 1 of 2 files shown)


sefroberg (changes requested):
Please check the version of Maven. 3.9.21 does not exist.

ubi10-openjdk-21.yaml
```diff
@@ -51,7 +51,7 @@ modules:
   - name: jboss.container.openjdk.jdk
     version: "21"
   - name: jboss.container.maven
-    version: "3.8.21"
+    version: "3.9.21"
```
sefroberg, line 54: Check your version 3.9.21 does not exist. It looks like the upstream head is 3.9.9

jmtd: 3.9.21 is the version of the cekit module (not the maven version specifically). It's used to select the right version when more than one is available, e.g. in the ubi9 branch you can see `modules/maven/11`, `modules/maven/17` and `modules/maven/21`. They all have the same module name, but different versions.

The scheme is `<maven  major version>.<maven minor version>.JDK version`.

jmtd: Belt and braces, the module and the image yaml match:
```
$ git grep 3.9.21
modules/maven/21/module.yaml:version: '3.9.21'
ubi10-openjdk-21.yaml:    version: "3.9.21"
```

sefroberg (approved):
Looks good to me.

PR #564, opened 2025-03-04
[OPENJDK-3647] adjust jlink  tests

jmtd:
Some behavioural change (S2I_ENABLE_JLINK is not echoed in the logs); fix syntax of final test; adjust approach used to check for the presence or absence of files and directories.

https://issues.redhat.com/browse/OPENJDK-3647


Josh-Matsuoka (approved):

sefroberg (approved):
Looks good to me.

PR #573, opened 2025-03-25
Bump container version for next release

jmtd:


jerboaa (commented):
Isn't JDK 11 deprecated?

sefroberg:
@jerboaa it is. @jmtd probably not needed.

sefroberg (approved):
I don't remember what we agreed for where to keep the version numbers for each image. If you want to keep the UBI 9 branch as the source of truth then this makes sense to update the jdk 11 images yamls. Otherwise I would leave that to the ELS branches.

jmtd:
Thanks @jerboaa @sefroberg for your catch; the ELS images should not rely on anything from this branch. I've removed bumping the version from the YAMLs and filed https://issues.redhat.com/browse/OPENJDK-3696 to track deleting the JDK11 image descriptors.

PR #577, opened 2025-04-11
    OPENJDK-3655  change script permissions to permit o+x

jmtd:
https://issues.redhat.com/browse/OPENJDK-3655

This sets o+x for executable scripts and removes x for files that are intended to be sourced but not directly executed.

As it stands this is as small change, but, my preferred approach to addressing this would be the approach taken in #559, which would result in a cleaner and more manageable source. @jerboaa , would you object to a fix like that in `ubi9`?


jerboaa (approved):
OK. GHA seems to have hit a timeout issue.

PR #579, opened 2025-04-16
[OPENJDK-3768] Update RHEL9 ELS containers for 2025-04 Security Update

vieiro:
This brings `jdk11-els-rhel9` on par with `ubi9`. Last commit also bumps ELS/RHEL9 image versions to `1.22`.


jmtd:
Hi @vieiro : this looks good, so long as your intention is to ship the CPU update and container feature updates in the same release. Is that right?

vieiro:
Will have to coordinate with SE for this, but that will be by next week! Let's close and retry then.

PR #580, opened 2025-04-16
[OPENJDK-3767] Update RHEL8 ELS containers for 2025-04 Security Update

vieiro:
This brings `jdk-11-els-rhel8` on par with `ubi8`. Last commit bumps els rhel8 image versions too.


jmtd:
The final, overall diff  (as github shows it anyway) has _only_ the version bump changes; this confuses me a little since there are ~7 other commits. I need to dig into that further Tomorrow. Also it moves the version beyond the upstream images which are still only on 1.21.

vieiro:
Let's retry this after Easter. Happy Easter!

PR #592, opened 2025-07-04
[OPENJDK-3935] Fix script permissions for artifacts

jmtd:
#583 caused an unintended break for images built on OSBS. For other build types, things worked as expected, but for OSBS builds the file permissions for copied-in files (which are round tripped through the dist-git lookaside cache) lacked +x permissions where they were needed.

This is a -- hopefully temporary -- fix that re-introduces some `configure.sh` scripts to force permissions, and some behave test to check that they applied correctly.

The longer-term fix is to be via https://github.com/cekit/cekit/issues/944

https://issues.redhat.com/browse/OPENJDK-3935

### DIFF
--- a/modules/jvm/configure.sh
+++ b/modules/jvm/configure.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+set -euo pipefail
+
+# OPENJDK-3695: force +x for scripts
+chmod 0755 /opt/jboss/container/java/jvm/*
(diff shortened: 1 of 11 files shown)


jmtd:
I want to add more behave tests to thoroughly check the relevant scripts are executable.

jerboaa (approved):

PR #594, opened 2025-07-23
[OPENJDK-4002] merge down the jlink-dev branch

jmtd:

This is a merge-down of the jlink-dev branch to ubi9. Rationale: to remove the need to continually sync the jlink-dev branch with ubi9  and ease rebuilding the TP image at the same time as the rest of them.

The jlink-dev branch modifies some core modules and the JDK21 image descriptor.

For this PR instead we create a new image descriptor. All the jlink changes to the regular modules are reverted and isolated to the jlink module and new TP image descriptor.

The exceptions are

 * introduction and use of "pathfinder" module. This abstracts the jar-finding logic out of `modules/run/artifacts/run-java.sh` so it can be re-used elsewhere.
 * the jlink code is invoked by additions to `modules/s2i/bash/artifacts/local/assemble`. These are entirely logic-gated by `if [ "$S2I_ENABLE_JLINK" = "true" ]`, so none of it will be executed by the normal images.
 * `modules/s2i/core/artifacts/s2i-core` adds an export of `S2I_ENABLE_JLINK` (defaulting to false) which should be benign


jmtd:
I appreciate 190 commits (so far) is a lot. I am keen to preserve the authorship of the jlink stuff for Josh and Jaya. Some useful comparison commands I used in preparing:

`git diff --stat ubi9 HEAD` (as well as inspecting the full diff)

`git config diff.colormoved plain` -- helps see chunks of code that are moved between files (for pathfinder in particular)

jmtd:
closed in favour of #596

PR #593, opened 2025-07-15
[OPENJDK-3975] tests for custom configuration and data

jmtd:
https://issues.redhat.com/browse/OPENJDK-3975

Adjust the remaining `rsync` calls to avoid trying to change permissions on the destination directories.

First, tests.

PR #595, opened 2025-07-31
[OPENJDK-4008] Revert artifacts changes that resulted in use of lookaside cache

jmtd:
https://issues.redhat.com/browse/OPENJDK-4008

The changes in [OPENJDK-3655](https://issues.redhat.com/browse/OPENJDK-3655) (#583) had the unintended side-effect of OSBS builds using the Brew lookaside cache to handle artifacts. We first noticed that this didn't preserve permissions properly (missing +x on scripts) which I initially worked around with some explicit `chmod`, but I've now also learned that, since this generates a `/sources` file, this also breaks the generation of source containers.

Revert the move to `artifacts:` entirely. Retain the tests we added to check file permissions, and  add some further changes (mostly deletions) to ensure the installed files have sensible permissions: +x for all where needed, and not where not; and all ownership is root:root.


jmtd:
For all but UBI9/JDK17, there is one failing test:
*   features/image/imagebasic.feature:20  Check installed scripts are executable by all users (OPENJDK-3655)
This is caused by https://github.com/rh-openjdk/openjdk-container-test-applications/pull/10 which was merged in anticipation of #593 

UBI9/JDK17 GHA run was terminated for unknown reasons as happens from time to time, rescheduling

jmtd:
That failed `  features/image/imagebasic.feature:20  Check installed scripts are executable by all users (OPENJDK-3655)` which it shouldn't have, investigating

jmtd:
Tests are good. The failign (error) test for the s2i images is due to the changes made to openjdk-container-test-applications for #593

jerboaa (approved):
This seems to revert the code changes from #583 without reverting the test. OK.

PR #596, opened 2025-08-05
[OPENJDK-4002] merge down jlink-dev branch (take 2)

jmtd:

This is a merge-down of the jlink-dev branch to ubi9. Rationale: to remove the need to continually sync the jlink-dev branch with ubi9  and ease rebuilding the TP image at the same time as the rest of them.

The jlink-dev branch originally modified some core modules and the JDK21 image descriptor. We now instead create a new image descriptor and all jlink changes are confined to that and the new jlink module, with the following exceptions:

1. introduction and use of "pathfinder" module. This abstracts the jar-finding logic out of `modules/run/artifacts/run-java.sh` so it can be re-used elsewhere.
2. the jlink code is invoked by additions to `modules/s2i/bash/artifacts/local/assemble`. These are entirely logic-gated by `if [ "$S2I_ENABLE_JLINK" = "true" ]`, so none of it will be executed by the normal images.
3. `modules/s2i/core/artifacts/s2i-core` adds an export of `S2I_ENABLE_JLINK` (defaulting to false) which should be benign


jerboaa (commented):
While I understand the rationale, this seems a risky move. I worry that this will break the UBI 9 image scripts in ways we don't anticipate. By keeping it separate we only risk breakage on the UBI9 tech-preview image that uses the jlink work. That's a much smaller set of people than the full UBI9 set of images.

Should we instead merge this into the UBI10 branch?

jmtd:
> While I understand the rationale, this seems a risky move. I worry that this will break the UBI 9 image scripts in ways we don't anticipate. By keeping it separate we only risk breakage on the UBI9 tech-preview image that uses the jlink work. That's a much smaller set of people than the full UBI9 set of images.

A less-invasive change would be to introduce the pathfinder module, but not refactor the existing modules to use it. That would still leave changes (2) and (3) above. What's your assessment of the risk of those?

> Should we instead merge this into the UBI10 branch?

That's an interesting idea.

At the moment, I've been updating and shipping the TP image  alongside the UBI9 containers. Merging this to UBI10 would possibly prevent me doing that, but we would still get automated respins for CVEs based on the last shipped version. That's probably fine.

jmtd:
closing in favour of  #610

PR #555, opened 2025-02-19
[OPENJDK-3661] don't set alternatives for runtime images

jmtd:
The method for setting alternatives in RHEL10 has changed (see RHEL-80074) but we don't (currently) need to set them anyway.
https://issues.redhat.com/browse/OPENJDK-3661

PR #605, opened 2025-09-03
sync ubi10 branch with ubi9 changes

jmtd:
This is a merge of recent changes to ubi9 into ubi10 and resolve conflicts.

Two new commits on top to resolve ubi10 specific problems

* e8f75eae492701a5c9dfc9f329b0953c7a23ac2d
* 76fb69ea57d834da5294c7b52d1e5f02c1805a4c

Note that this will introduce the fixlabel GHA action for ubi10 going forward. So I will have to file JIRAs for branch-sync merges after this one.


jmtd:
This has regressed #555 . I'll push a fix for that here.

jmtd:
runtime test stalled on  `Scenario: Check that directories from other JDKs are not present (JDK21)  # features/jboss.container.openjdk/openjdk.feature:31` with the periodic problem we have. Rescheduling

jmtd:
@jerboaa this is going to be necessary every now and then until UBI10 goes GA. I'm wondering whether we waive reviews for merges from ubi9 to ubi10. I'm not sure since they're never clean (as this PR shows). I don't want to lean on your time too much. WDYT?

modules/maven/s2i/tests/features/files.feature
```diff
@@ -0,0 +1,10 @@
+Feature: test file properties for Maven S2I module
+
+  # builder-only test
+  @ubi9/openjdk-17
+  @ubi9/openjdk-21
```
jerboaa: Do those tags need an update? for `@ubi10`?

jerboaa (approved):
Looks OK. One question about the `@ubiXX` tags.

modules/maven/s2i/tests/features/files.feature
jmtd: Yes they do, thanks for catching!

PR #607, opened 2025-09-04
[OPENJDK-3666] merge jboss.container.java.s2i.{core,bash} modules

jmtd:
https://issues.redhat.com/browse/OPENJDK-3666


jmtd:
the two test failures are fixed in #609

jmtd:
@jfdenise can I request you TAL this and assess it from the perspective of impact on wildfly? TIA!

modules/maven/s2i/artifacts/opt/jboss/container/maven/s2i/maven-s2i
```diff
@@ -1,7 +1,7 @@
 
 # include dependencies
 source "${JBOSS_CONTAINER_UTIL_LOGGING_MODULE}/logging.sh"
-source "${JBOSS_CONTAINER_S2I_CORE_MODULE}/s2i-core"
```
jfdenise, line 4: There is an implicit dependency on the `jboss.container.java.s2i` module.
We can't depend on `jboss.container.java.s2i` module that implements the s2i logic, WildFly has its own different s2i logic.
We need the core to be clean from java.s2i specifics.

jmtd:
Hi @jfdenise , thank you for taking the time to look at this.
Do I understand correctly:
* you depend on `jboss.container.s2i.core`
* `jboss.container.s2i.core` is kind of an "API" and library-like module, rather than direct logic -- you use bits of it and so do we
* the "driver" logic in our case is `jboss.container.java.s2i.bash`, and you don't and cannot use this module

If my understanding is correct I should not merge them. I would like to perhaps move/rename them to make it clearer that one is generic and one is java (us) specific.

jfdenise:
@jmtd , I fully agree with your proposal to rename to make it clearer. Thank-you!

jmtd:
I started work on this but renaming means changing a lot of references everywhere.  I think I've run out of time for this at the moment, so I'll close.

PR #609, opened 2025-09-04
[OPENJDK-3653] fix s2i path in tests

jmtd:
The new location is /usr/libexec/s2i

https://issues.redhat.com/browse/OPENJDK-3653


jerboaa (approved):
Thanks for fixing it!

PR #610, opened 2025-09-04
4002 ubi10 merge jlink

jmtd:

Following the suggestion in #596 , this is a merge-down of the `jlink-dev` branch targetting ubi10.

I've aimed to preserve the commit history of the work (and the attribution), so there are a huge number of commits and various views onto the changes are useless.

The clearest diff view (for me) is `git diff 397d0f4 eeb56296aa6bf4018e04e221e379d1e8e3e19e80`, i.e. from the merge base to the HEAD of the PR branch. The changes made to _existing_ images are

* `run-java.sh` requires and includes the `pathfinder` module, to where we've moved the logic to detect various types of jar file in the deployments directory, 
* the s2i module now defines S2I_ENABLE_JLINK defaulting to false
* s2i/assemble is extended to call the jlink scripts if  S2I_ENABLE_JLINK=true

THe PR adds a new image, openjdk-21-jlink-rhel9, which is unusual as it's based on UBI9 in this branch rather than UBI10.

This does _not_ enable the jlink functionality in the UBI10 images -- but -- that's something we should discuss separately.


jerboaa:
@jmtd Could we merge latesest ubi10 branch into this. Which should make the labelling issue go away. Also have we considered adding a JDK 25 based GHA check now that JDK 25 is installable via the ubi10 repos?

jmtd:
@jerboaa : 
> Could we merge latesest ubi10 branch into this.

Sure. I'll push that now. In doing so it looks like commits from #607 snuck in (last merge), which shouldn't be there, so I'll resolve that too.

> Also have we considered adding a JDK 25 based GHA check now that JDK 25 is installable via the ubi10 repos?

I'll raise a JIRA, we should definitely do this. Edit: [OPENJDK-4195](https://issues.redhat.com/browse/OPENJDK-4195) 	Add GHA checks for JDK25 images

jerboaa:
> @jerboaa :
> 
> > Could we merge latesest ubi10 branch into this.
> 
> Sure. I'll push that now. In doing so it looks like commits from #607 snuck in (last merge), which shouldn't be there, so I'll resolve that too.

Am I missing something? I don't see a merge from recently or a GHA re-run....

jmtd:
> Am I missing something? I don't see a merge from recently or a GHA re-run....

Seems I forgot to push it? Sorry. I need to get those unwanted commits out from #607 before it can be looked at again anyway. I'll untag you as reviewer until I get that sorted. Sorry!

jmtd:
I've force-pushed a fresh merge onto the current `ubi10` branch, to ensure we don't pick up #607 and to resolve the GHA labelling action as well.

modules/jlink/artifacts/opt/jboss/container/java/jlink/mkjreimage.sh
```diff
@@ -0,0 +1,15 @@
...
+depsfile="$S2I_JLINK_TEMP_PATH/module-deps.txt"
+
+function generate_jre_image() {
+	test -f $depsfile
+	modules="$(cat $depsfile)"
+
+	$JAVA_HOME/bin/jlink --output "$S2I_JLINK_OUTPUT_PATH" \
+    	   	--add-modules "$modules" \
+		--strip-debug --no-header-files --no-man-pages \
+		--compress=2
```
jerboaa:
```suggestion
		--compress=zip-6
```

From JDK 21 `jlink --help`:
```
      --compress <compress>             Compression to use in compressing resources:
                                        Accepted values are:
                                        zip-[0-9], where zip-0 provides no compression,
                                        and zip-9 provides the best compression.
                                        Default is zip-6.
                                        Deprecated values to be removed in a future release:
                                        0:  No compression. Equivalent to zip-0.
                                        1:  Constant String Sharing
                                        2:  Equivalent to zip-6.
```

modules/jlink/artifacts/opt/jboss/container/java/jlink/preflight.sh
```diff
@@ -0,0 +1,25 @@
...
+
+jlink_preflight_check()
+{
+    # preflight check: do we have what we need?
+    if [ "$JAVA_VERSION" -lt 11 ]; then
+      echo "Jlink integration not available for JDK${JAVA_VERSION}!"
+      echo "Jlink integration is only supported for JDK versions 11 and newer."
+      exit 1
+    fi
+    if [ ! -d /usr/lib/jvm/java/jmods ]; then
```
jerboaa:
```suggestion
    if [ ! -d $JAVA_HOME/jmods ]; then
```

This would otherwise possibly check for a different JDK that is also installed. Unlikely but a possiblity.

modules/jvm/tests/features/memory.feature
```diff
@@ -34,6 +34,7 @@ Feature: OPENJDK-559 JVM Memory tests
 
   # Not the runtime images
   @ubi10/openjdk-21
+  @openjdk-tech-preview/openjdk-21-jlink-rhel9
```
jerboaa, line 37: We are going to merge into `ubi10` branch and it mentions `...-rhel9`.

templates/jlink/jlinked-app.yaml
```diff
@@ -0,0 +1,307 @@
...
+    labels:
+      app: ${APPNAME}
+      app.kubernetes.io/part-of: ${APPNAME}
+  spec:
+    lookupPolicy:
+      local: true
+    tags:
+    - from:
+        kind: DockerImage
+        name: registry.access.redhat.com/ubi9/ubi-micro
```
jerboaa, line 169: Should we adapt this to UBI 10?

jerboaa (commented):
Please consider the following comments:
- Changes to `run-java.sh` would also affect any non-jlink-on-the-fly use-cases, correct? So there is some chance for breakage.
- Please check tagging of tests. They have `openjdk-21-jlink-rhel9`. Mentioning `rhel-9` when this is a `ubi10` branch PR.
- Some references are referring to UBI 9, I wonder if we should adapt it to ubi 10. Is the expectation to build the UBI 9-jlink preview from the ubi10 branch?

jmtd:
Thank you for such a thorough review!

> Changes to run-java.sh would also affect any non-jlink-on-the-fly use-cases, correct? So there is some chance for breakage.

Yes, there is that risk. It's mitigated to some  extent by the UBI10 images being a new "product", and so not going to break existing customers or deployments. The extent of the changes to `run-java.sh` are the refactoring of code into `pathfinder` that we wished to re-use elsewhere in the jlink scripts. If you have git config `diff.colormoved=plain`, it's possible to have machine-assistance to see that _most_ of the delta in `run-java.sh` is moved as-is into pathfinder, if that helps to guage the size of the change/risk.

(If I was doing the jlink project again today, I would think carefully about doing a refactor like _pathfinder_ again. On the other hand, I think it does improve the quality of the code.)

The same concern applies to other scripts: `/usr/local/s2i/assemble` has additional code for jlink which could theoretically cause problems for non-jlink usage in any of the builders (e.g. a syntax error in the scripts causing termination). The mitigation there was to try to ensure all the changes were within the branch on `if [ "$S2I_ENABLE_JLINK" = "true" ]; then`; looking again at it now, we could also move the following lines within that branching (I think I will push this change if you don't object):

```
+# run the pathfinder scripts to define JAVA_APP_JAR and JAVA_LIB_DIR
+source "${JBOSS_CONTAINER_UTIL_PATHFINDER_MODULE}/pathfinder.sh"
+setup_java_app_and_lib
```

> Some references are referring to UBI 9, I wonder if we should adapt it to ubi 10. Is the expectation to build the UBI 9-jlink preview from the ubi10 branch?

We should do some thinking (IMHO separately to, and after, this PR)  about what to do about the tech-preview image, and the jlink feature in general, going forward. Do we merge the feature into all the images? Has it had sufficient exposure? We've had zero feedback, but I have had one SA tell me their customer was trying it out. So that was good to know.

This PR does not attempt to upgrade the jlink TP image from rhel9 to rhel10. It is odd, and contrasting, to have it (as a rhel9-based image) in the ubi10 branch. The rationale is to try and keep the jlink-dev code in a live branch, so it does not bitrot in the gap before we can revisit it. Otherwise the cost of catching it up with other image changes may be very big.

We could upgrade the TP image to ubi10, but we have not agreed to ship a ubi10-based jlink TP image, and (for the time being at least) we have been shipping CVE fixes for the ubi9-based TP image (we have not committed to doing that, but it has come along "for free" with the other images so far).

jmtd:
the jlink image descriptor doesn't build (due to relying upon the deleted tzdata module); I will restore that.

jerboaa (approved):
OK. The only comment I have is to add a comment to the relevant files for `tzdata`. I.e. if we end up removing the jlink TP then we can remove that again.

PR #612, opened 2025-09-30
[OPENJDK-4112] Add JDK and JRE 25 cekit modules

jmtd:
https://issues.redhat.com/browse/OPENJDK-4112

so we can install JDK25 (once available)

### DIFF
--- a/modules/jdk/25/module.yaml
+++ b/modules/jdk/25/module.yaml
@@ -1,0 +1,32 @@
+schema_version: 1
+
+name: "jboss.container.openjdk.jdk"
+description: "Installs the JDK for OpenJDK 25."
+version: &jdkver "25"
+
+labels:
+- name: "org.jboss.product"
+  value: "openjdk"
+- name: "org.jboss.product.version"
+  value: *jdkver
+- name: "org.jboss.product.openjdk.version"
+  value: *jdkver
+
+envs:
+- name: "JAVA_HOME"
+  value: "/usr/lib/jvm/java-25"
+- name: "JAVA_VENDOR"
+  value: "openjdk"
+- name: "JAVA_VERSION"
+  value: *jdkver
+- name: JBOSS_CONTAINER_OPENJDK_JDK_MODULE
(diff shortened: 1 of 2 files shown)


modules/jdk/25/configure.sh
```diff
@@ -0,0 +1,11 @@
...
+set -euo pipefail
+
+# Update securerandom.source for quicker starts (must be done after removing jdk 8, or it will hit the wrong files)
+JAVA_SECURITY_FILE=/usr/lib/jvm/java/conf/security/java.security
+SECURERANDOM=securerandom.source
+if grep -q "^$SECURERANDOM=.*" $JAVA_SECURITY_FILE; then
+    sed -i "s|^$SECURERANDOM=.*|$SECURERANDOM=file:/dev/urandom|" $JAVA_SECURITY_FILE
+else
+    echo $SECURERANDOM=file:/dev/urandom >> $JAVA_SECURITY_FILE
+fi
```
jerboaa: Unfortunately, this after-the-fact modification of `java.security` will break the JEP-493-style `jlink` from the runtime image. The same could be achieved with adding `-Djava.security.egd=file:/dev/urandom` to the java launch command which doesn't have that problem. Has this been considered as an alternative approach to this?

jmtd: Thanks for pointing that out. I hadn't considered doing it the way you suggest but see no reason not to do so, especially to un-break JEP-493 use.  I want to look again at whether we should do this at all ([OPENJDK-4113](https://issues.redhat.com/browse/OPENJDK-4113)). And assuming we should, if it's to be applied across all JDK versions, I would like to move the configuration away from the JDK-specific paths (perhaps to `jboss.container.java.jvm`)

jmtd: It should just go: #613

jerboaa: Agreed.

jerboaa (approved):
LGTM

PR #613, opened 2025-10-02
OPENJDK-4113: stop adjusting securerandom.source

jmtd:
Since kernel v5.6 (2020), /dev/random no longer blocks and is functionally equivalent to /dev/urandom. So there's  no point in adjusting securerandom.source.

[OPENJDK-4113](https://issues.redhat.com/browse/OPENJDK-4113)


jmtd:
Should we consider this for ubi9 too?

jerboaa (approved):
OK. This should be in some form of release note and mentioning the `-Djava.security.egd=file:/dev/urandom` workaround. Note that while newer kernels treat them, `/dev/random` and `/dev/urandom`, equivalently, we don't know on which host system the image will run. It might be RHEL-9, or ... The container compatibility matrix from OCP might come in handy here.

jerboaa:
> Should we consider this for ubi9 too?

No. Similar to other "larger" changes. Let's leave ubi9 alone as much as possible.

jerboaa:
Aside: Can we disable the label/JIRA issue action until it's fixed?

jmtd:
> Aside: Can we disable the label/JIRA issue action until it's fixed?

Sure. I've no idea why it's broken. It worked for about a day, and then stopped.

PR #615, opened 2025-10-07
[OPENJDK-4136] clean up FROM in overrides

jmtd:
UBI9 base image not appropriate for UBI10

https://issues.redhat.com/browse/OPENJDK-4136****


jerboaa (commented):
Content sets in the same files don't seem to be correct either.

jmtd:
> Content sets in the same files don't seem to be correct either.

Yeah. I will probably delete them. They won't be needed (in this place) for Konflux builds. I haven't got as far as dealing with content sets under Konflux yet, so I left alone.

jerboaa (approved):

jerboaa:
> > Content sets in the same files don't seem to be correct either.
> 
> Yeah. I will probably delete them. They won't be needed (in this place) for Konflux builds. I haven't got as far as dealing with content sets under Konflux yet, so I left alone.

OK.

PR #618, opened 2025-11-20
JDK25 images for UBI9

jmtd:
Top-level epic: https://issues.redhat.com/browse/OPENJDK-4185

(rewritten 2025-12-09 after refactoring and force-pushing [7a8eb1b](https://github.com/rh-openjdk/redhat-openjdk-containers/pull/618/commits/7a8eb1bef5ebc00f12a4f1e124cf93448cd623de))

This introduces JDK25 images for UBI9.

At the time of writing, JDK25 is not available in RHEL9. Therefore, this also introduces jdk/jre modules with version `latest`, which use `java-latest-openjdk`, and depend upon another new module `epel` which sets up the EPEL repository.

The redhat overrides files override the image's `jdk` module version to `25`, which also removes the transitive dependency on the `epel` module, so that does not go into redhat builds. (I have a test build I can share details about privately.)

All other matters aside, would be prepared to merge this given the above situation?


modules/jdk/latest/configure.sh
```diff
@@ -0,0 +1,24 @@
...
+alternatives --set jre_openjdk java-25-openjdk.${_arch}
+
+# Update securerandom.source for quicker starts (must be done after removing jdk 8, or it will hit the wrong files)
+JAVA_SECURITY_FILE=/usr/lib/jvm/java/conf/security/java.security
+SECURERANDOM=securerandom.source
+if grep -q "^$SECURERANDOM=.*" $JAVA_SECURITY_FILE; then
+    sed -i "s|^$SECURERANDOM=.*|$SECURERANDOM=file:/dev/urandom|" $JAVA_SECURITY_FILE
+else
+    echo $SECURERANDOM=file:/dev/urandom >> $JAVA_SECURITY_FILE
+fi
```
jerboaa: I thought we don't need this anymore?

jmtd: We don't; that fix was done in `ubi10` and is yet to be backported to `ubi9`. I'll remove it from at least the new modules as part of this PR.

jmtd:
The failing tests for JDK25 (as of [94a17b9](https://github.com/rh-openjdk/redhat-openjdk-containers/pull/618/commits/94a17b9397d12d19c4d891701d15975bbbc04af4)) are

```
2025-11-27T15:50:16.4693986Z Errored scenarios:
2025-11-27T15:50:16.4694558Z   features/jboss.container.openjdk/openjdk.feature:36  Check that only OpenJDK 25 is installed
2025-11-27T15:50:16.4695611Z   features/jboss.container.openjdk/openjdk.feature:94  Check that directories from other JDKs are not present (JDK25)
```

The tests detect JDK21 -- this is pulled in by `maven-openjdk21`, which will be replaced with `maven-openjdk25` once it is available. This is a useful check that the test is working.

jmtd:
(note to self): I'm working on the non-EPEL version, and I've set up the JDK25 module in a similar way to the ubi10 approach: JDK25 depending upon JRE25 leaning on it to provide the common configuration, including the jvm-options file. I have not backported the same change for the other JDK versions. However this breaks because `modules/jvm/artifacts/opt/jboss/container/java/jvm/java-default-options` expects the path to be `"${JBOSS_CONTAINER_OPENJDK_JDK_MODULE}/jvm-options"`. For that path to be consistent across JDKs I would either need to duplicate jvm-options to the jdk25 module or backport the module de-duplication for the other JDK versions. I am leaning toward the latter.

jmtd:
Several GHA tests timed out, I am scheduling re-runs.

jmtd:
I haven't enabled `singleton-jdk` for these images, although it is necessary at the moment. Unfortunately it will not work with the EPEL/"latest" provided JDK (it will incorrectly assume the correct RPM name is `java-25-openjdk` based on `jdkver). It also shouldn't be necessary for RHEL, once a few other things are put into place.

redhat/ubi9-openjdk-25-runtime.yaml
```diff
@@ -0,0 +1,32 @@
+osbs:
+  configuration:
+    container:
+      compose:
+        pulp_repos: true
+        packages:
+        - java-25-openjdk-headless
+        signing_intent: unsigned
```
jerboaa:
```suggestion
        signing_intent: release
```

redhat/ubi9-openjdk-25.yaml
```diff
@@ -0,0 +1,34 @@
+osbs:
+  configuration:
+    container:
+      compose:
+        pulp_repos: true
+        packages:
+        - java-25-openjdk
+        - java-25-openjdk-devel
+        - java-25-openjdk-headless
+        signing_intent: unsigned
```
jerboaa:
```suggestion
        signing_intent: release
```

jerboaa (approved):
Looks good.

PR #620, opened 2025-11-28
[OPENJDK-4284] Shrink JDK modules, re-use env vars from JRE modules

jmtd:
In the JRE modules, set JAVA_HOME to a path which is valid for both the builder and runtime images.

In the JDK modules, stop defining JAVA_HOME, JAVA_VENDOR and JAVA_VERSION: inherit the values from the JRE modules instead.

https://issues.redhat.com/browse/OPENJDK-4284


jerboaa:
Please merge latest `ubi10` branch to see JDK 25 testing as well. Thanks!

jerboaa (approved):

PR #624, opened 2025-12-10
Use singleton-jdk for JDK25 images

jmtd:
It's not clear yet that we will have `maven-openjdk25` available at the time we want to ship. So, we may need to pull `maven-openjdk21` in the build phase and clean up multiple JDKs.


jerboaa (approved):
seems fine.

jmtd:
EPEL-based build fails due to this commit, but in a really strange place; however, redhat-overrides builds are fine.

jmtd:
JDK25 test here is stalling, but it passed for me locally.

PR #632, opened 2026-03-02
[OPENJDK-4569] reference correct maven module version

jmtd:
https://issues.redhat.com/browse/OPENJDK-4569

### DIFF
--- a/ubi10-openjdk-25.yaml
+++ b/ubi10-openjdk-25.yaml
@@ -56,9 +56,8 @@ modules:
   - name: jboss.container.openjdk.jdk
     version: "25"
   - name: jboss.container.maven
-    version: "3.9.21"
+    version: "3.9.25"
   - name: jboss.container.java.s2i.bash
-  - name: jboss.container.java.singleton-jdk
 
 help:
   add: true


ubi10-openjdk-25.yaml
```diff
@@ -56,9 +56,8 @@ modules:
   - name: jboss.container.openjdk.jdk
     version: "25"
   - name: jboss.container.maven
-    version: "3.9.21"
+    version: "3.9.25"
   - name: jboss.container.java.s2i.bash
-  - name: jboss.container.java.singleton-jdk
```
jerboaa, line 61: Any particular reason why we don't need this?

jerboaa (commented):
One question:

ubi10-openjdk-25.yaml
jmtd: Yeah. When we were pulling in the older maven version, it was installing `maven-openjdk21` which depends upon `java-21`, which we cleaned up with `singleton-jdk`. With 3.9.25, it installs `maven-openjdk25`.

jerboaa: OK, thanks!

jerboaa (approved):
OK

jerboaa:
Still surprised by the GHA failures.

jmtd:
> Still surprised by the GHA failures.

Me too :(

PR #635, opened 2026-03-05
[OPENJDK-4600] Replace redhat/* with konflux metadata

jmtd:
The data in redhat/* is relevant only for OSBS builds, which we cannot use for UBI10 images. Replace it with data that is useful for Konflux builds instead.

The metadata is not valid for upstream cekit: it's used by my konflux branch https://github.com/jmtd/cekit/tree/konflux

https://issues.redhat.com/browse/OPENJDK-4600

### DIFF
--- a/redhat/ubi10-openjdk-21-runtime.yaml
+++ b/redhat/ubi10-openjdk-21-runtime.yaml
@@ -1,27 +1,12 @@
-osbs:
-  configuration:
-    container:
-      compose:
-        pulp_repos: true
-        packages:
-        - java-21-openjdk-headless
-        signing_intent: release
+konflux:
   repository:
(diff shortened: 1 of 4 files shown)


jmtd:
Hm, the konflux dev branch of cekit doesn't like this stuff being in overrides.

jerboaa (approved):
OK. Though I'm not a fan of those internal leakages.

jmtd:
> OK. Though I'm not a fan of those internal leakages.

Me neither. We (and cekit) avoided it with the OSBS workflow because cekit invoked `rhpkg` (or `fedpkg`) behind the scenes and it encoded the final URIs. We unfortunately don't have the same mechanism available to us here.

PR #637, opened 2026-03-13
[OPENJDK-4598] stop using pkg-update module

jmtd:
This is a no-op with hermetic builds
https://issues.redhat.com/browse/OPENJDK-4598


jerboaa (approved):
OK. We may need to reconsider further down the line.

PR #646, opened 2026-06-30
Add env file for ubi9/openjdk-25-runtime

jmtd:
This is consumed by the IBM DeliveryPlatform.


jmtd:
Some more context. The DeliveryPlatform supports (or is going to support) Cekit natively. The way to pass arguments to the CI cekit is via an env file, such as the one in this PR. Merging this will mean we can start testing the Cekit pipeline for OpenJDK images, starting with ubi9/openjdk-25-runtime.

jerboaa (commented):
Could this move to `.cekit/openjdk-25-runtime-ubi9.env` perhaps?

jmtd:
> Could this move to .cekit/openjdk-25-runtime-ubi9.env perhaps?

Yes, the path is configurable. The file is more related to the DeliveryPlatform (which reads it and converts it into a cekit invocation) than cekit itself (which doesn't natively supoort an env file). If the intention is to reduce clutter in the root, what do you think of `ibm/openjdk-25-runtime-ubi9.env` (analogous to `redhat/*`, which exists for redhat-specific build quirks)?

jmtd:
@jerboaa I've moved the file to a new `ibm/` subdir, mirroring `redhat/`. WDYT?

